PR #27, opened 2026-02-21
fix: implement cosign signature inheritance for OCI index children 

### LINKED ISSUE #17: Cosign signed status not shown for child artifacts in OCI index depth view
## Summary

When a multi-arch (OCI index / manifest list) image is signed with cosign, the parent manifest list correctly shows the green checkmark in the "Signed" column, but all child platform images show a red X (unsigned) when viewed through the OCI index depth view.

This affects both scenarios:
- Signing the **manifest list** by tag (signature attached to the index)
- Signing a **child image** directly by digest

## Steps to Reproduce

1. Push a multi-arch image to Harbor:
   ```bash
   crane copy docker.io/library/alpine:latest localhost:8080/cosign-test/signed-image:alpine --insecure
   ```

2. Sign the manifest list with cosign:
   ```bash
   cosign sign --key cosign.key --allow-insecure-registry --allow-http-registry \
     localhost:8080/cosign-test/signed-image:alpine
   ```

3. Verify the signature:
   ```bash
   cosign verify --key cosign.pub --allow-insecure-registry --allow-http-registry \
     localhost:8080/cosign-test/signed-image:alpine
   # Verification succeeds
   ```

4. Open the Harbor UI → Projects → cosign-test → signed-image → Artifacts tab
   - The manifest list shows **green checkmark** in the Signed column ✅

5. Click the OCI index icon to drill into child platform images
   - **All child images show red X** (unsigned) ❌

## Expected Behavior

Child artifacts within a signed OCI index should either:
- Inherit the signed status from their parent manifest list, or
- At minimum, display an indicator that the parent index is signed

## Actual Behavior

- Parent manifest list: Green checkmark (correct)
- Child images (linux/amd64, linux/arm64, etc.): Red X (misleading)

### Parent manifest list — green checkmark
Parent shows signed

### Child images in depth view — all red X
Children show unsigned

## API Confirmation

The API confirms the signature accessory is only attached to the parent, not propagated to children:

```bash
# Parent — has cosign accessory
curl ".../artifacts/sha256:25109184...?with_accessory=true"
# → accessories: [{"type": "signature.cosign", ...}]

# Child — no accessories
curl ".../artifacts/sha256:1529d135...?with_accessory=true"
# → accessories: null
```

This is technically correct (cosign signs exactly the digest it targets), but the UX is misleading since users expect the signed status to be visible on child images when the parent index is signed.

## Customer Report Context

A customer on v2.13.2 reported this exact scenario: they signed an image, `cosign verify` succeeds, but the Harbor UI shows red X. Their `cosign verify` output confirmed the signature was on a specific child digest within a manifest list.

Related upstream issues (different root cause — cosign v3 bundle format):
- https://github.com/goharbor/harbor/issues/22401
- https://github.com/goharbor/harbor/issues/22592
- https://github.com/goharbor/harbor/issues/22643
- https://github.com/goharbor/harbor/pull/22628

## Environment

- Harbor: latest main branch (post-v2.14.1)
- cosign: v3.0.4 (also reproduced with v2.2.4)
- Image: multi-arch `alpine:latest` (all platforms)

Aloui-Ikram:
before : 
image

image

After : 
image
image

---
## Summary by cubic
Implements cosign signature inheritance from OCI index parents so unsigned children show as signed when the parent is signed. Updates the accessories API to surface inherited signatures when the UI filters for cosign.

- **Bug Fixes**
  - Inherit a parent's cosign signature when a child has none by scanning parent references and attaching the signature for display.
  - ListAccessories returns inherited cosign signatures on "signature.cosign" queries and fixes a local variable name shadowing issue.

- **Refactors**
  - Add Artifact.InheritedAccessories and populate it in populateAccessories to keep inherited signatures separate from direct accessories.
  - Fetch artifacts with accessories via artCtl.Get(..., WithAccessory: true) and combine direct + inherited when serving API results.

<sup>Written for commit eb03f8f48788a69bb4dae9a74049268094b0abf0. Summary will update on new commits.</sup>

 Fixes #17 

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying inherited signatures from parent artifacts in artifact listing views.
  * Enhanced signature detection to automatically discover and propagate signatures from parent artifacts when direct signatures are unavailable.

### DIFF
--- a/src/controller/artifact/controller.go
+++ b/src/controller/artifact/controller.go
@@ -783,44 +783,47 @@ func (c *controller) populateAdditionLinks(ctx context.Context, artifact *Artifa
 }
 
 func (c *controller) populateAccessories(ctx context.Context, art *Artifact) {
-        accs, err := c.accessoryMgr.List(ctx, q.New(q.KeyWords{"SubjectArtifactID": art.ID}))
-        if err != nil {
-                log.Errorf("failed to list accessories of artifact %d: %v", art.ID, err)
-                return
-        }
-        art.Accessories = accs
+	accs, err := c.accessoryMgr.List(ctx, q.New(q.KeyWords{"SubjectArtifactID": art.ID}))
+	if err != nil {
+		log.Errorf("failed to list accessories of artifact %d: %v", art.ID, err)
+		return
+	}
+	art.Accessories = accs
 
-        // Check for inherited Cosign signatures from parent OCI Index
-        hasCosign := false
-        for _, acc := range accs {
-                if acc.GetData().Type == accessorymodel.TypeCosignSignature {
-                        hasCosign = true
-                        break
-                }
-        }
+	// Check for inherited Cosign signatures from parent OCI Index
+	hasCosign := false
+	for _, acc := range accs {
+		if acc.GetData().Type == accessorymodel.TypeCosignSignature {
+			hasCosign = true
+			break
+		}
+	}
 
-        // If the child is not signed, look up its parents to see if they are signed
-        if !hasCosign {
-                parents, err := c.artMgr.ListReferences(ctx, &q.Query{
-                        Keywords: map[string]any{"ChildID": art.ID},
-                })
-                if err != nil || len(parents) == 0 {
-                        return
-                }
-                for _, p := range parents {
-                        parentAccs, err := c.accessoryMgr.List(ctx, q.New(q.KeyWords{"SubjectArtifactID": p.ParentID}))
-                        if err != nil {
-                                continue
-                        }
-                        for _, pAcc := range parentAccs {
-                                if pAcc.GetData().Type == accessorymodel.TypeCosignSignature {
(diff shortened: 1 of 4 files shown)


bupd (approved):
/lgtm

bupd (changes requested):
@Aloui-Ikram this has one issue - the way you handle the inherited is wrong - it should not say signed in standalone artifacts. 

simply, inheritance of sign should only apply when the artifact is exclusively reachable as a child of the signed artifact, 
it should not apply when the artifact is independently pushed as standalone 

Image

bupd:
i am merging this one - after deep diving in - there is a way these things are handled. and the standalone artifact showing is basically the verbose way of harbor showing the artifacts - it has to do with the way i created the multi-arch image - since i pushed the childs first and the index afterwards - this happnd - anyways its good now that we know how this would behave in different scenarios. so yeah.

Screenshot_2026-03-18_05-54-48

PR #33, opened 2026-03-02
feat: Add Conditional Immutability Rules Compatible with Retention Policy

bupd:
Extends immutability rules to support conditional matching using retention policy rule templates, making immutability work analogously to retention.

- Add `Parameters` field to `immutable.Metadata` model so immutability rules can carry rule-template parameters (e.g. number of days)
- Update the immutability matcher to evaluate retention rule templates (`nDaysSinceLastPush`, `nDaysSinceLastPull`, `latestPushedK`, `latestPulledN`, `immutable_template`) as conditions before marking a tag immutable
- Expose five immutable rule templates via the retention metadata endpoint: always, pushed within last N days, pulled within last N days, most recently pushed # artifacts, most recently pulled # artifacts
- Refactor `retention/policy/action/performer` into its own sub-package to break the import cycle introduced by sharing rule evaluators
- Update portal UI (`add-immutable-rule` component) to render the condition selector, add `CallbackPipe` for dynamic template rendering, and wire up i18n strings across all supported locales
- Fix `isImmutable` to fail-closed on matcher error (was fail-open, could allow deletion of protected artifacts)
- Fix `immutable_template` alias normalization applied consistently in both `Get()` and `Valid()`
- Fix duplicate rule detection to compare params, not just template and selectors
- Fix merged text rendering in immutable tag rule list (missing space between action and template spans)

### New Features

- **Conditional immutability rules**: Tags can now be protected based on activity conditions, not just pattern matching. Supported conditions:
  - Always (existing behavior)
  - Pushed within the last N days
  - Pulled within the last N days
  - Most recently pushed # artifacts
  - Most recently pulled # artifacts
- All five conditions reuse the existing retention rule evaluators - no new backend logic required.
- Full i18n support across 9 locales (de, es, fr, ko, pt-BR, tr, zh-CN, zh-TW).

### Bug Fixes

- `isImmutable` now fails closed on matcher error - artifacts are protected rather than deleted when the immutability check encounters an error (e.g. DB connectivity issue).
- `immutable_template` backward-compat alias is now normalized consistently in `Valid()` as well as `Get()`.

## Upstream PR (Waiting)
- https://github.com/goharbor/harbor/pull/22047

---
## Summary by cubic
Adds conditional immutability rules that reuse retention templates (always, pulled/pushed within N days) so projects can lock tags based on activity. Updates API and UI to support rule parameters and condition selection, and refactors the retention performer to share evaluators safely.

## Release Notes

* **New Features**
  * Added support for conditional immutable tag rules - users can now apply immutability based on "always", time-based conditions (pulled/pushed within X days), or count-based conditions (most recently pushed/pulled # artifacts)
  * Enhanced rule creation UI with condition selector and configurable parameters

* **Internationalization**
  * Added full multi-language support for immutable rules in German, Spanish, French, Korean, Portuguese, Turkish, Simplified Chinese, and Traditional Chinese

* **Improvements**
  * Extended retention policy framework to support template-based conditional rules


Vad1mo:
the existing tag retention system also supports two additional **count-based** templates that users are asking for:

| Template | Condition |
|---|---|
| `latestPushedK` | Retain the most recently **pushed** # artifacts |
| `latestPulledN` | Retain the most recently **pulled** # artifacts |

**Request:** Could you also expose `latestPushedK` and `latestPulledN` as immutability condition templates?

**Use case from users:**
- *David Caro:* "Is there something like 'latest X pushed artifacts'? It might be that we don't push or pull in many days, but still want to keep say the 5 latest artifacts immutable."
- *Jonas Beck:* "It would be great if it would support the same settings as tag retention allows — retain the most recently pushed/pulled # artifacts."

bupd:
@Vad1mo added your suggestions

PR #99, opened 2026-04-04
ci: Split test workflow into parallel pure and DB jobs

### LINKED ISSUE #98: CI: Optimize Test workflow from 26min to ~10min
The Test workflow currently takes **~26 minutes** ([run #23916839235](https://github.com/container-registry/harbor-next/actions/runs/23916839235)). Most of that time is avoidable.

### Timing Breakdown

| Step | Duration | Notes |
|------|----------|-------|
| Set up job | 3s | |
| Checkout | 2s | |
| Setup Go | **89s** | 1.2GB module cache download from GitHub API at ~22 MB/s + extract |
| Setup Task | 2s | |
| Start PostgreSQL/Redis | **41s** | `apt-get update && install build-essential postgresql redis-server` every run |
| Generate API code | 19s | `go install swagger` + swagger generate |
| Run unit tests | **23m 15s** | 557 packages, `-p 1 -parallel 1` (fully serial) |

### 1. Use GitHub Actions `services:` instead of `start-ci-services.sh` (~41s saved)

The `start-ci-services.sh` script runs `apt-get install postgresql redis-server` on every job. GitHub Actions has native `services:` support that starts containers **in parallel with job setup** — they're ready before the first step runs.

```yaml
services:
  postgresql:
    image: postgres:16-alpine
    env:
      POSTGRES_USER: postgres
      POSTGRES_PASSWORD: root123
      POSTGRES_DB: registry
    ports:
      - 5432:5432
    options: >-
      --health-cmd pg_isready
      --health-interval 10s
      --health-timeout 5s
      --health-retries 5
  redis:
    image: redis:7-alpine
    ports:
      - 6379:6379
    options: >-
      --health-cmd "redis-cli ping"
      --health-interval 10s
      --health-timeout 5s
      --health-retries 5
```

- Works on both `ubuntu-latest` and self-hosted runners (DinD sidecar already configured)
- Eliminates `apt-get install` entirely
- Removes `build-essential` dependency from test path
- `max_connections = 300` can be set via a small post-start step or custom entrypoint

### 2. Split tests into parallel pure + serial DB jobs (~12min saved)

Tests run `-p 1 -parallel 1` because "many DAO suites share the same test database." But only **~87 out of 316** test packages actually touch the database. The other **~229 are pure unit tests** with no DB dependency.

Proposal: two parallel jobs:

- **`test-unit`**: Pure packages (no `lib/orm`, `dao`, or `POSTGRESQL` imports) → run with `-p 4` or higher
- **`test-db`**: DB-dependent packages → keep `-p 1 -parallel 1`

Both jobs run in parallel. The serial DB tests take ~11min, pure tests with `-p 4` take ~3-5min. Wall time drops from 23min to ~11min.

A script can auto-classify packages:

```bash
# Packages that import orm/dao → serial DB tests
db_pkgs=$(go list ./... | xargs -I{} sh -c \
  'go list -f "{{.ImportPath}} {{.Imports}}" {} | grep -qE "lib/orm|/dao" && echo {}')

# Everything else → parallel pure tests
pure_pkgs=$(go list ./... | grep -vFf <(echo "$db_pkgs"))
```

### 3. Pre-install PostgreSQL + Redis in self-hosted runner image

The custom runner image (`runner-image/Dockerfile`) can include `postgresql` and `redis-server`. The `start-ci-services.sh` script already guards with `command -v` checks and skips `apt-get` if present. This is a fallback optimization for when `services:` can't be used.

## Expected Result

| | Current | Optimized |
|---|---------|-----------|
| Service startup | 41s (apt-get) | 0s (services: pre-started) |
| Unit tests | 23m (serial) | ~11m (parallel split) |
| **Total** | **~26 min** | **~13 min** |

## References

- Test workflow run analyzed: https://github.com/container-registry/harbor-next/actions/runs/23916839235
- `start-ci-services.sh`: `.github/scripts/start-ci-services.sh`
- Test workflow: `.github/workflows/test.yml`

Vad1mo:
## Summary

Optimizes the test workflow from ~26 min to ~13 min by splitting tests into two parallel CI jobs and running lint/vuln checks concurrently.

- **test-unit**: 498 pure packages with full parallelism (no DB/Redis needed)
- **test-db**: 59 DB-dependent packages running serially with `services:` containers (replaces `apt-get install postgresql redis-server`)
- Taskfile `test:ci` runs lint:api, lint:report, vuln-report in parallel with unit tests via `deps:`
- New `task test:unit:pure` and `task test:unit:db` tasks for targeted local testing
- Local runs skip `-count=1` for Go test caching; CI always uses `-count=1`
- Auto-classifies packages by grepping test files for `PrepareTestForPostgresSQL`/`GiveMeRedisPool`

### Timing breakdown (estimated)

| | Current (single job) | Optimized (parallel jobs) |
|---|---|---|
| Service startup | 41s (apt-get) | 0s (services: pre-started) |
| Pure tests | — (all serial) | ~5 min (parallel, own runner) |
| DB tests | — (all serial) | ~8 min (serial, own runner) |
| **Wall time** | **~26 min** | **~13 min** |

## Related Issues

Closes #98

## Type of Change

## Testing
- [x] Package classification verified: 498 pure, 59 DB, 557 total (after exclusions)
- [x] `list-test-packages.sh` tested locally — runs in <0.2s
- [x] `task test:unit:pure` runs successfully with full parallelism
- [x] YAML syntax validated

## Checklist

## Release Notes

* **Chores**
  * Split test suite into separate jobs for unit tests and database-dependent tests, enabling parallel execution and improved CI efficiency.
  * Automated test package categorization to distinguish tests requiring external services.
  * Enhanced test coverage tracking with separate reports for each test category.


Vad1mo:
The `list-test-packages.sh` script works but has two weaknesses: it's grep-based (no compiler enforcement) and new DB tests could silently end up in the wrong bucket. Two complementary changes would eliminate the script entirely:

### 1. Build tags for DB tests (`//go:build db`)

Move `PrepareTestForPostgresSQL` behind a build tag so the compiler enforces correct classification:

```go
// common/dao/testutils_db.go
//go:build db

package dao

func PrepareTestForPostgresSQL() {
    // ... existing implementation
}
```

Any test file calling `dao.PrepareTestForPostgresSQL()` without `//go:build db` on itself **fails to compile** — the function doesn't exist in the default build. This is self-enforcing: you can't forget the tag.

CI becomes:
```bash
# Job 1: pure tests (full parallelism, no services needed)
go test ./...

# Job 2: DB tests (serial, with services)
go test -tags db -p 1 -parallel 1 ./...
```

~66 test files need `//go:build db` added. `list-test-packages.sh` is no longer needed.

### 2. Graceful skip when DB is unavailable

Change the `log.Fatalf` in `PrepareTestForPostgresSQL` and `InitDatabaseFromEnv` to `os.Exit(0)` when `POSTGRESQL_HOST` is unset (same for `GiveMeRedisPool` with `REDIS_HOST`):

```go
func PrepareTestForPostgresSQL() {
    dbHost := os.Getenv("POSTGRESQL_HOST")
    if len(dbHost) == 0 {
-       log.Fatalf("environment variable POSTGRESQL_HOST is not set")
+       fmt.Println("SKIP: POSTGRESQL_HOST not set, skipping DB tests")
+       os.Exit(0) // TestMain exit 0 = package passes with no tests run
    }
    // ... rest unchanged
}
```

This improves local dev ergonomics — `go test ./...` just works everywhere, DB tests gracefully skip when no database is available. Combined with build tags, `go test -tags db ./...` still skips cleanly if you haven't started services.

### Combined effect

- **Compiler-enforced classification** — no runtime script, no grep, no silent miscategorization
- **`go test ./...` always works** — DB tests either excluded (no `-tags db`) or skip gracefully
- **`list-test-packages.sh` deleted** — classification moves from a shell script to the Go type system

Vad1mo:
**Correction on point 2 (graceful skip):** The `os.Exit(0)` approach has a footgun — if `POSTGRESQL_HOST` is set in your shell (from dev environment, `.env`, etc.), running `go test ./...` would silently include DB tests at full parallelism, corrupting test data. You'd have to remember to unset it.

Build tags alone (point 1) don't have this problem. DB tests are excluded at compile time unless you explicitly pass `-tags db`. No env var to manage.

The graceful skip would only help the edge case of running `go test -tags db ./...` without a database available. Not worth the risk. **Recommend build tags only.**

PR #155, opened 2026-04-15
perf(ci): Cache Go modules, swagger binary, and generated API code

bupd:
## Summary
- Cache the swagger binary (`~/go/bin/swagger`, keyed on `versions.env`) and generated API code (`src/server/v2.0/` + `.task/checksum`, keyed on `swagger.yaml` + templates) via `actions/cache` — these aren't on the host-mounted `$GOMODCACHE`/`$GOCACHE` so they were regenerated every run (~2 min overhead per job)
- Switch `gen-apis` to `method: checksum` so Task uses content hashing instead of timestamps — timestamps break after cache restore, checksums don't
- Add clarifying comments to `setup-go-cached` explaining that self-hosted runners use host-mounted caches
- Also includes the shard index fix for DinD container name collisions (#154)

## How it works (steady state, cache hit)
1. Host-mounted `$GOMODCACHE`/`$GOCACHE` handle Go module and build caching (already in place)
2. `actions/cache` restores `~/go/bin/swagger` → `require-swagger` status check passes, skip install
3. `actions/cache` restores `src/server/v2.0/` + `.task/checksum/` → `gen-apis` checksum matches, skip codegen
4. Tests start with near-zero overhead

## Type of Change

## Testing
- [x] CI will validate — first run populates caches, subsequent runs should skip gen-apis entirely

## Checklist

### DIFF
--- a/.github/actions/setup-go-cached/action.yml
+++ b/.github/actions/setup-go-cached/action.yml
@@ -18,7 +18,8 @@ runs:
         go-version-file: ${{ inputs.go-version-file }}
         cache: false
 
-    # GitHub-hosted only: restore module cache from GitHub's cache storage
+    # GitHub-hosted only: restore module cache from GitHub's cache storage.
+    # Self-hosted runners use persistent host-mounted $GOMODCACHE instead.
     - name: Restore Go module cache
       if: runner.environment != 'self-hosted'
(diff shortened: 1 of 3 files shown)


bupd:
there is not really a big difference with and without this - its just the same - closing this 

This is actually slower than what we have.

PR #208, opened 2026-05-12
ci: Compose Release Artifacts With Task

bupd:
## Summary
- Move release artifact build/push/sign/SBOM work into composable Taskfile release tasks.
- Keep release-please and release note generation in GitHub Actions where GitHub context is required.
- Pin release tooling versions for release-please-action, cosign, and syft.

## Testing
- task --list | rg 'release:'
- task --summary release:run
- python -c 'import yaml; yaml.safe_load(open(".github/workflows/release-please.yml"))'
- git diff --check


bupd:
closing things as why change when things already work you know. main reason is we have better release pipeline much more integrated with taskfile - so closing this.

PR #212, opened 2026-05-13
fix: Cache the Scannability Lookups Per Artifact-List Request

Vad1mo:
## Summary

- `GET /api/v2.0/projects/{p}/repositories/{r}/artifacts` becomes very slow (TTFB > 60s, replication clients see nginx 504) on repositories holding many top-level OCI image-index artifacts whose children are not directly scannable by the configured scanner.
- Caches `scan.checker.isAccessory` and `HasUnscannableLayer` results on the per-request `checker` so shared descendants of multiple index roots are looked up at most once per response — instead of once per root.
- On a 38-index response where each index referenced ~180 children, this drops the assembler stage from **~24s / ~13,800 SQL** to **~1.1s / ~540 SQL**; handler-equivalent TTFB drops from **~24s to ~1.4s**.

## Problem

https://github.com/container-registry/8gcr/issues/136

### Observed behaviour

Real-world chain of events seen on Harbor 2.14.1 behind a default nginx-ingress:

1. A second Harbor instance runs a `type=harbor` pull-mode replication policy against this Harbor.
2. The destination jobservice calls `…/artifacts?page=1&page_size=100&with_tag=…&with_scan_overview=…&with_signature=…`.
3. harbor-core takes ~54s to assemble the response.
4. nginx-ingress hits the 60s `proxy_read_timeout` because harbor-core writes the response only after the assembler finishes (no early write), so the whole 54s counts as a gap between successive reads.
5. nginx closes the upstream socket → harbor-core logs `Handler crashed with error write tcp …: write: broken pipe`.
6. nginx returns `504 Gateway Time-out` to the replication client.
7. jobservice retries; each retry repeats the timeout. The replication execution fails after several retries with:

   ```
   failed to fetch artifacts: failed to list artifacts of repository
   '<project>/<repo>': http error: code 504, message <html><head><title>504 Gateway Time-out
   ```

The slow peer link is not the cause — a fast client running the same call from inside the same datacenter sees TTFB ≈ 54s.

### Reproducer (cURL)

```bash
curl -sS -u admin:'<password>' \
  -o /dev/null \
  -w 'HTTP %{http_code} ttfb=%{time_starttransfer}s size=%{size_download}\n' \
  'https://<harbor-host>/api/v2.0/projects/<project>/repositories/<repo>/artifacts?page=1&page_size=100&with_tag=true&with_label=true&with_scan_overview=true&with_signature=true&with_immutable_status=true&with_accessory=true'
```

If `ttfb` approaches the ingress `proxy-read-timeout` (default 60s), replication clients will hit `504` and retry-loop until the execution fails.

### Per-artifact cost on the affected instance

| Page size | Artifacts in response | Body size | TTFB (fast client) |
| --------- | --------------------- | --------- | ------------------ |
| 100       | 38                    | ~3.2 MB   | **~54 s**          |
| 10        | 10                    | ~860 KB   | ~15 s              |

Roughly **~1.4s of server compute per artifact** with the default `with_*` flags. The cost scaled linearly with the number of top-level artifacts in the response — not with the size of the underlying repository — because the dominant cost is per-artifact assembly, not the row scan.

### Root cause

`assembler.ScanReportAssembler.Assemble` (`src/server/v2.0/handler/assembler/report.go`) calls `scan.checker.IsScannable` once per response artifact. For an OCI image index, `IsScannable` runs `artifactCtl.Walk` over every descendant, and the walk callback runs two DB queries per node:

- `accessory.Mgr.List({ArtifactID, Digest})` to skip cosign / SBOM / attestation accessories
- `artifactCtl.HasUnscannableLayer(digest)` to skip in-toto SBOMs masquerading as images

When the response holds many image-index roots that share the same child digests — which is normal for fat indexes carrying many platform/variant manifests — those same descendants get walked once per root, because `Walk`'s de-duplication map is local to a single `Walk` call. There is no cross-call cache on the `checker`.

Measured on an affected database (38 index-typed top-level artifacts, ~180 children each, all children sharing the same set of digests across parents):

| Stage | Queries | Time |
| --- | ---: | ---: |
| `controller.List` (cheap part) | 191 | ~0.2s |
| `ScanReportAssembler.Assemble` (the slow part) | **13,792** | **~24s** |
| Total handler-equivalent TTFB | | **~24s** |

That's ~363 SQL queries per top-level artifact = 1 + 180 children × 2 queries each, repeated 38× because the same children are re-queried for every parent.

In production over an RDS connection with concurrent traffic the same workload pushes past the 60s ingress budget, which is when replication breaks.

## Fix

`src/controller/scan/checker.go` now keeps two per-request caches on the `checker` struct:

- `accessoryCache map[int64]bool` keyed by artifact ID
- `unscannableLayerCache map[string]bool` keyed by digest

`assembler.NewScanReportAssembler` creates a fresh `scan.NewChecker()` per request, so the cache lifetime is implicitly one HTTP request. `Assemble` iterates artifacts sequentially within that lifetime, so no concurrent map access happens. `hasCapability` is already in-memory and doesn't need a cache.

After the fix, on the same 38-index response:

| Stage | Queries | Time |
| --- | ---: | ---: |
| `controller.List` | 191 | ~0.2s |
| `ScanReportAssembler.Assemble` (batch — same call shape the handler runs) | **536** | **~1.1s** |
| Total handler-equivalent TTFB | | **~1.4s** |

~26× fewer queries, ~17× lower TTFB. Well below the 60s ingress budget.

## Regression test

`src/server/v2.0/handler/assembler/slow_list_repro_test.go` (build tag `db`) drives `artifact.Ctl.List` + `ScanReportAssembler.Assemble` against a database the operator points at via `SLOW_REPRO_REPO=<project>/<repo>`. The test:

- Logs SQL count and wall time for the controller list, per-artifact assembler, and batch assembler stages so future regressions show up in the test log.
- Fails if handler-equivalent TTFB exceeds 30s — half the standard 60s ingress budget.

The test is opt-in (it skips when `SLOW_REPRO_REPO` is unset) because reproducing the slow path requires data with many shared-descendant image indexes. Easiest way to seed such data synthetically is to push N image-index artifacts each referencing the same M child manifests.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Performance / hot-path optimisation

## Release Notes

Fix slow `GET /api/v2.0/projects/{p}/repositories/{r}/artifacts` (and the resulting nginx 504s on Harbor-to-Harbor replication) for repositories that hold many top-level OCI image-index artifacts whose child manifests are shared and not directly scannable. The artifact-list path now reuses the scannability lookup across roots within a single request.

## Testing

- [x] `go test -count=1 ./src/controller/scan/...` (unit, unchanged behaviour)
- [x] `go test -count=1 ./src/server/v2.0/handler/assembler/...` (unit)
- [x] `SLOW_REPRO_REPO=<repo> go test -tags db -run TestSlowArtifactListRepro ./src/server/v2.0/handler/assembler/...` against a database that exhibits the bug — before fix: ~24s TTFB / ~13,800 SQL, after fix: ~1.4s TTFB / ~540 SQL.

## Checklist

- [x] PR title follows conventional commits (`fix: …`)
- [x] No new lint warnings (`go vet` clean)
- [x] Generated code unchanged (no swagger edits)

### DIFF
--- a/src/controller/scan/checker.go
+++ b/src/controller/scan/checker.go
@@ -35,10 +35,12 @@ type Checker interface {
 // NewChecker returns checker
 func NewChecker() Checker {
 	return &checker{
-		artifactCtl:   artifact.Ctl,
-		accMgr:        accessory.Mgr,
-		scannerCtl:    scanner.DefaultController,
-		registrations: map[int64]*models.Registration{},
+		artifactCtl:           artifact.Ctl,
+		accMgr:                accessory.Mgr,
+		scannerCtl:            scanner.DefaultController,
+		registrations:         map[int64]*models.Registration{},
+		accessoryCache:        map[int64]bool{},
+		unscannableLayerCache: map[string]bool{},
 	}
 }
 
@@ -47,6 +49,18 @@ type checker struct {
 	accMgr        accessory.Manager
 	scannerCtl    scanner.Controller
 	registrations map[int64]*models.Registration
+	// accessoryCache caches isAccessory results by artifact ID for the lifetime
+	// of this checker. The artifact list handler creates a fresh checker per
+	// request via assembler.NewScanReportAssembler, so the cache is implicitly
+	// request-scoped. Avoids requerying the same descendant across many
+	// IsScannable calls in a single response — the worst case is an artifact
+	// list where most rows are OCI image indexes whose child manifests are
+	// shared (so the same children would otherwise be looked up once per root).
+	accessoryCache map[int64]bool
+	// unscannableLayerCache caches HasUnscannableLayer results keyed by digest
+	// for the same reason: a single digest can appear under many parents in one
+	// list response.
+	unscannableLayerCache map[string]bool
 }
 
 func (c *checker) IsScannable(ctx context.Context, art *artifact.Artifact) (bool, error) {
@@ -91,7 +105,7 @@ func (c *checker) IsScannable(ctx context.Context, art *artifact.Artifact) (bool
 		// when scanning these type of sbom artifact, the scanner might assume it is image layer with tgz format, and if scanner read the layer with a stream of tgz,
 		// it fail and close the stream abruptly and cause the pannic in the harbor core log
 		// to avoid pannic, skip scan the in-toto sbom artifact sbom artifact
-		unscannable, err := c.artifactCtl.HasUnscannableLayer(ctx, a.Digest)
+		unscannable, err := c.hasUnscannableLayer(ctx, a.Digest)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


bupd:
Dump-backed verification against the linked customer database confirms the performance claim.

Page size: `100`  
Artifacts returned: `38`

| Measurement | Before / control | After fix | Improvement |
|---|---:|---:|---:|
| `controller.List` | 191 SQL, ~0.2s | unchanged | not the bottleneck |
| Scan assembly SQL | 13,866 SQL | 536 SQL | ~25.9x fewer SQL queries |
| Scan assembly time | ~30.1s | ~1.0s | ~29.8x faster |
| Handler-equivalent TTFB | ~30.3s | ~1.2s | ~24.7x faster |

This validates that the request-scoped scannability cache removes the repeated descendant lookups across shared OCI image-index children and brings the handler-shaped artifact-list path well below the ingress timeout budget.

PR #232, opened 2026-05-21
fix(security): Validate blob-mount source project and reject tokens missing iat

Vad1mo:
Follow-up to #31 (and upstream goharbor/harbor#22900, upstreamed in goharbor/harbor#23270).

Addresses two review findings on the `tokenIssuedAfterProjectCreation` check introduced in #31:

### 1. Blob-mount source project is not validated (auth bypass)
The check only looked at `info.ProjectName`. A cross-repository blob mount — `POST /v2/<repo>/blobs/uploads/?from=<src>` — additionally requires **pull** access to the source repository (`accessList` in `server/middleware/v2auth/access.go`), and `artifactinfo` middleware populates `ArtifactInfo.BlobMountProjectName`. The source project's creation time was never checked, so a token issued before the *source* project was deleted and recreated with the same name was still accepted — re-opening, for the source project, the exact hole the function closes for the destination.

Fix: validate `iat` against both `ProjectName` and (when set and different) `BlobMountProjectName`.

### 2. Missing `iat` claim panics the request handler
`claims.IssuedAt` is a `*jwt.NumericDate` and was dereferenced unconditionally (`claims.IssuedAt.Time`). The validator does not require `iat`, so a signed token without `iat` reached the dereference and panicked. Fail closed: reject tokens without `iat`.

### Tests
- `TestTokenIssuedAfterProjectCreation_NilIAT` — token without `iat` is rejected, no panic.
- `TestTokenIssuedAfterProjectCreation_BlobMountSource` — destination created before `iat`, source recreated after `iat` ⇒ rejected.

Both new tests fail on the pre-fix code (the nil-`iat` case panics) and pass after the change. Existing table-driven cases unchanged. Verified with `go test -tags db ./server/middleware/security/`.

### DIFF
--- a/src/server/middleware/security/v2_token.go
+++ b/src/server/middleware/security/v2_token.go
@@ -79,22 +79,40 @@ func (vt *v2Token) Generate(req *http.Request) security.Context {
 }
 
 // tokenIssuedAfterProjectCreation prevents tokens from a deleted project
-// granting access to a new project recreated with the same name.
+// granting access to a new project recreated with the same name. It validates
+// every project the request is authorized against, including the source
+// project of a cross-repository blob mount.
 func tokenIssuedAfterProjectCreation(ctx context.Context, logger *log.Logger, claims *v2TokenClaims) bool {
-	info := lib.GetArtifactInfo(ctx)
-	if info.ProjectName == "" {
-		return true
-	}
-	p, err := project_ctl.Ctl.GetByName(ctx, info.ProjectName)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


Vad1mo:
This PR it is already merged upstream https://github.com/goharbor/harbor/pull/23270 and backported to 2.15 https://github.com/goharbor/harbor/pull/23297

Vad1mo:
Close if we get it from upstrem sync

bupd:
closing since this is already cherry-picked.

PR #318, opened 2026-06-16
perf(core): Avoid eager structured-logger build on demoted auth-failure logs

Vad1mo:
Backport of #317 to `release-2.15` (follow-up to #314).

On the auth-failure path the basic-auth generator built the structured logger (`log.WithField(...).WithField(...)`) unconditionally, then logged at DEBUG for the expected bad-credential case. Harbor's `WithField` → `WithFields` eagerly clones the logger, copies the field map, and `fmt.Sprintf`-formats each field regardless of level, while `Debugf` no-ops at INFO — so every bad-credential probe paid the full structured-logger cost for a discarded line. The `WithField` chain is now built only on the ERROR branch; the bad-credential DEBUG path logs plainly. ERROR behavior is unchanged.

## Type of Change

- [x] Performance / hardening (non-breaking)

## Release Notes

_None — internal logging micro-optimization on the auth-failure path._

## Testing

- [x] `gofmt`, `go build`, `go vet` clean on `server/middleware/security`

### DIFF
--- a/src/server/middleware/security/basic_auth.go
+++ b/src/server/middleware/security/basic_auth.go
@@ -82,13 +82,13 @@ func (b *basicAuth) Generate(req *http.Request) security.Context {
 	})
 
 	if err != nil {
-		logEntry := log.WithField("client IP", GetClientIP(req)).WithField("user agent", GetUserAgent(req))
-		// Bad credentials are expected probe traffic (scanners/bots) -> DEBUG;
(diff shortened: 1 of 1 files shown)


Vad1mo:
Closing — the release-2.15 backport is handled automatically via `/backport v2.15` once #317 merges. No need for a manual PR.

PR #397, opened 2026-06-30
perf(blob): Skip redundant blob Touch when blob is already StatusNone

BojanZelic-DD:
## Summary

On every `PUT manifest`, `probeBlob` calls `Touch` on the manifest blob even when the row is already `StatusNone`. `Touch` is a single-row `UPDATE blob SET version = version + 1, update_time = now(), status = 'none' ...` that takes a row lock. 

When the same content digest is re-pushed concurrently (CI fan-out, reproducible rebuilds, multi-arch, retries), those updates serialize on the same row and block `PUT manifest`. This was causing P99 latency spikes for us sometimes in the case of 10+ minutes for a PUT manifest operation to complete. 

When querying the database we saw manifest blob rows with `version` of 250–580 (Touched that many times/day), all `status = none` — every one a redundant `none → none` write contending on the lock.

`Touch` exists only to coordinate with GC, so this gates it to the cases that need it:
- `StatusNone` → skip the redundant write; only refresh `update_time` when stale enough (older than half the GC time window) that a not-yet-associated blob could fall inside the GC candidate window.
- `StatusDelete` / `StatusDeleteFailed` → still rescue (a GC candidate the push needs).
- `StatusDeleting` → unchanged.

No change to push behavior; GC race protection is preserved. Works whether GC is enabled or not.

This is the P99 latency of the PUT Manifest operation before and after the change;
image

## Type of Change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Documentation (`docs:`)
- [ ] Tests (`test:`)

## Release Notes

Fixed `PUT manifest` latency spikes (up to 10+ minutes at P99) caused by row-lock contention on the `blob` table. Pushes and blob HEAD requests no longer issue a redundant status `UPDATE` when the blob is already in the normal state, which removes the hot-row contention seen under concurrent re-pushes of the same content digest. GC race protection is unchanged, including when `GC_TIME_WINDOW_HOURS` is set to zero.

## Testing

## Checklist


Vad1mo:
@BojanZelic-DD thanks, we will look into it

Vad1mo (commented):
I traced the mechanics end to end; confirming the analysis and noting one correctness gap I fixed directly on the branch.

## Why the impact is minutes, not milliseconds

The redundant `UPDATE` alone wouldn't produce 10-minute P99s. The multiplier is that **PUT manifest runs inside a request-wide DB transaction**: `transaction.Middleware` (`src/core/middlewares/middlewares.go`) wraps the whole request, and PUT manifest is not in `dbTxSkippers` (PATCH/PUT blob upload and GET/HEAD are). `probeBlob` runs in the *before* hook, so the row lock acquired by `Touch` is held across the proxied storage write and all after-hooks. Concurrent re-pushes of the same digest therefore serialize their **entire requests**. It's also why the identical unconditional `Touch` in `head_blob.go` never showed up in your graphs — HEAD skips the transaction middleware and runs autocommit.

On a steady-state re-push the after-hooks write nothing to the blob row (`Ensure` returns early, `Sync` only writes on content-type change), so the probe `Touch` was the only hot-row write — skipping it removes the contention completely for this case.

I verified the half-window heuristic is GC-safe: candidates require `update_time` older than the full window (`GetBlobsNotRefedByProjectBlob`), the skip only applies when it's younger than half, and any `Touch` between mark and sweep bumps `version`, which fails sweep's `version = ?` CAS.

## The gap: `window <= 0`

With `GC_TIME_WINDOW_HOURS=0` (used in GC testing/e2e), the `window > 0 &&` guard inverted the safe behavior: `StatusNone` was then *never* touched while every unreferenced blob is an instant GC candidate — mark and sweep CAS both succeed against the unchanged `version`, and GC can delete a blob mid-push. The old unconditional Touch rescued exactly this. A non-positive window has no safety margin, so it must mean *always* touch.

## Pushed to this branch (3f006c61b)

- Fixed the polarity: `window <= 0 || time.Since(bb.UpdateTime) > window/2`, extracted as `shouldTouchNone` in `util.go`.
- Applied the same gating to `head_blob.go` — no latency win there (autocommit), but it's the bulk of the redundant `none → none` write churn (every layer HEAD on every push) and keeps the two paths consistent.
- Tests: unit table for `shouldTouchNone` (incl. zero/negative window) and db-tagged suites asserting `version` is unchanged on fresh re-push/HEAD and bumped when stale or `StatusDelete`.

Also retitled the PR to Conventional-Commits form (`perf: ...`) since we squash-merge the title, and aligned the release-notes figure with the body.

One residual worth a follow-up issue (not this PR): the *necessary* touches (rescue/stale) still run inside the request-wide transaction, so those still hold the lock for the whole request. Rare, but the same failure mode.

Vad1mo:
@BojanZelic-DD as you see in my addtion we have applied this also the head requests. Can you give it a spin and verify it solves your problme, we can go in and merg it in.

BojanZelic-DD:
@Vad1mo I think your change makes sense. During the period of high p99 latency for PUT v2_manifest we also saw higher P99 latency for HEAD v2_blob operations (although not as severe ~ 2-3s vs normally it's about 200ms - 1s)

BojanZelic:
@Vad1mo thoughts on merging this?

bupd (approved):
/lgtm

bupd:
@BojanZelic-DD Thanks for your contribution 

the merging of this needed other improvements to the other parts too.

PR #406, opened 2026-07-02
upstream: Add ECDSA Certificate Support For Internal TLS

### LINKED ISSUE #405: ECDSA certificates for internal tls aren't supported
**Expected behavior and actual behavior:**

When using cert-manager with ECDSA certificates (e.g., P-256) for internal TLS CA, Harbor's token service fails to parse ECDSA private keys with the error:

```
x509: failed to parse private key (use ParseECPrivateKey instead for this key format)
```

This prevents components from authenticating and pulling images. The issue manifests as `500 Internal Server Error` when the registry service attempts to generate tokens for image pulls or authentication.

**Steps to reproduce the problem:**

1. Use cert-manager to generate an ECDSA-based internal TLS CA with `algorithm: ECDSA` and `size: 256`
2. Configure Harbor to use this ECDSA certificate for internal TLS (e.g., in Kubernetes cert-manager setup)
3. Attempt to pull a public image or login with `docker login`

**Expected result:** Image pull or login succeeds using ECDSA-signed certificates

**Actual result:** `500 Internal Server Error` from token service; token generation fails with "x509: failed to parse private key"

**Versions:**

- harbor version: 2.10.0+

**Related Issues:**

- Upstream: goharbor/harbor#20082

rossigee:
## Summary

Enables Harbor to use ECDSA certificates (P-256, P-384, P-521) for internal TLS, resolving token generation failures when cert-manager generates ECDSA-based CA certificates.

The token service now automatically detects key type (RSA vs ECDSA) and selects the appropriate JWT signing algorithm (RS256 vs ES256/384/512). Supports both SEC1 (`EC PRIVATE KEY`) and PKCS#8 (`PRIVATE KEY`) PEM formats for ECDSA keys, and maintains backward compatibility with existing RSA configurations.

Fixes #405

## Type of Change
- [x] Upstream Harbor cherry-pick (`upstream:` goharbor/harbor#22861)
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## Release Notes

Adds support for ECDSA certificates in Harbor's internal TLS token signing. Harbor now automatically detects whether the configured certificate uses RSA or ECDSA and selects the appropriate JWT signing algorithm (RS256 for RSA, ES256/384/512 for ECDSA). This enables users to deploy Harbor with cert-manager-generated ECDSA certificates for internal component communication.

## Testing

- [x] Manual testing performed (ECDSA token generation and parsing with P-256, P-384, P-521 curves; SEC1 and PKCS#8 formats)
- [x] Existing tests pass (backward compatibility with RSA configurations verified)

## Checklist

- [x] Tests updated/added


rossigee:
Thank you for the thorough reviews! I'll address both issues:

### 1. PKCS8 Support in MakeToken (CodeRabbit)
You're correct—`MakeToken()` in `authutils.go` re-parses the PEM with `libtrust.UnmarshalPrivateKeyPEM()`, which doesn't support PKCS8 format. I'll update it to derive the `kid` from the already-parsed key in `NewOptions` instead, avoiding the re-parsing issue.

### 2. Unsafe Type Assertion in Tests (Cubic AI)
Good catch on the test isolation issue. I'll switch from `assert.IsType` (which doesn't halt) to `require.IsType` to fail fast on type mismatches.

Working on fixes now.

src/core/service/token/test/ecdsa_private_key.pem
Vad1mo, line 1: Can we get away not checking in keys? this will make all vuln checks go crazy forever.  Even if it is for test. 
take a look in our taskfile setup there are flow that we use to generate keys easily .

bupd: @rossigee

PR #408, opened 2026-07-02
upstream: Allow Database Authentication When OIDC Is Enabled For Non-OIDC Users

### LINKED ISSUE #407: Allow database authentication when OIDC is enabled for non-OIDC users
**Expected behavior:**

When Harbor is configured with OIDC authentication enabled, users who were created in Harbor's local database before OIDC was enabled should still be able to authenticate using their database credentials.

**Actual behavior:**

Non-OIDC users (users without OIDC metadata) are locked out when OIDC authentication is the configured auth mode, preventing them from logging in or performing operations.

**Steps to reproduce:**

1. Create a local database user in Harbor
2. Enable OIDC authentication in Harbor's configuration
3. Attempt to login with the locally-created user's credentials
4. Login fails even though the user exists in the database

**Expected result:** Login succeeds using database credentials as fallback

**Actual result:** Login fails; user is rejected because they lack OIDC metadata

**Rationale:**

During OIDC migration, there may be existing local users (e.g., administrative accounts) that need to continue working. This change enables a fallback to database authentication for users without OIDC metadata, preventing accidental lockout during the migration process.

**Related Issues:**

- Upstream: goharbor/harbor#23416

rossigee:
## Summary

Enables local database authentication as a fallback for users without OIDC metadata when OIDC is configured as the primary authentication method. This prevents accidental lockout of existing local users (e.g., admin accounts) during OIDC migration.

The login flow now checks if a user has OIDC metadata; if not, it falls back to database authentication instead of rejecting them outright.

Fixes #407

## Type of Change
- [x] Upstream Harbor cherry-pick (`upstream:` goharbor/harbor#23416)
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## Release Notes

Adds fallback database authentication for users without OIDC metadata when OIDC is enabled. This allows existing local users to continue authenticating with their database credentials during OIDC migration, preventing lockout of administrative or legacy accounts.

## Testing

- [x] Unit tests added for all canUseOIDCAuth code paths (DB error, user not found, no OIDC metadata, has OIDC metadata)
- [x] Unit tests verify IsSuperUser behavior with mocked user manager
- [x] Existing tests pass (backward compatibility verified)

## Checklist

- [x] Tests updated/added

### DIFF
--- a/src/core/auth/auth_test.go
+++ b/src/core/auth/auth_test.go
@@ -15,13 +15,16 @@ package auth
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 
 	"github.com/goharbor/harbor/src/common/models"
+	"github.com/goharbor/harbor/src/pkg/user"
 	"github.com/goharbor/harbor/src/pkg/usergroup/model"
+	testinguserpkg "github.com/goharbor/harbor/src/testing/pkg/user"
 )
 
 var l = NewUserLock(2 * time.Second)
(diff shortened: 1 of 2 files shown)


Vad1mo (commented):
I do not understand why the PR admin users can always log in with username/password even when OIDC is enabled and in primary mode.

Vad1mo:
We need to rethink the usefulness of this PR, this was never an issue in all the years, and represents a very specific edge case, that is opening up quite a bit of confusion and explanation

bupd:
@rossigee I dont think this PR is needed - harbor-next already does this out of the box - you could have normal db users eventhough you manage your users via OIDC. 

only thing to note here is that you need to do it manually via updating in the DB.

Vad1mo:
closing it for now

rossigee:
> only thing to note here is that you need to do it manually via updating in the DB.

that was kind of the point of this PR - if I have OIDC enabled and I've misconfigured it, some details change and/or the service goes down, i don't want to have to mess around in SQL to regain access.

rossigee:
Thanks for taking a look. I understand the hesitation — this has worked in practice for most setups because the superuser fast-path (UserID == 1) has masked the issue for primary admins. But the problem is real for non-superuser local accounts. Here's what the source shows:

In `Login()` (`src/core/auth/authenticator.go:147`), when `auth_mode` is `oidc_auth`, the function uses `registry["oidc_auth"]` directly. The OIDC backend doesn't implement `Authenticate()` — it inherits the default which returns `ErrNotSupported`. There's no fallback to DB auth in that path.

The only exceptions are robot accounts and the superuser check (`IsSuperUser`), which is hardcoded to `UserID == 1`. So any other local user (admin or otherwise) created before OIDC migration can't log in with their DB credentials once OIDC is enabled.

This is a real lockout risk during OIDC migration, especially if the OIDC provider goes down or is misconfigured. The fix simply checks whether the user has OIDC metadata before deciding which backend to use — it doesn't change behavior for OIDC users at all.

For reference, the upstream PR is [goharbor/harbor#23416](https://github.com/goharbor/harbor/pull/23416).

PR #410, opened 2026-07-02
upstream: Prevent Silent HTTP 201 On Manifest Blob Association Failures

rossigee:
## Summary

Adds defensive error recovery to manifest push operations when artifact registration fails after a successful proxy response (HTTP 201). By verifying the ResponseBuffer middleware wrapper is in place, the handler can detect and recover from failures downstream, preventing storage-only orphans.

## Type of Change
- [x] Upstream Harbor cherry-pick (`upstream:` goharbor/harbor#23375)
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## Release Notes

Fixes a bug where manifest pushes that succeeded at the proxy level (HTTP 201) but failed during artifact registration would incorrectly return 201 to the client, leaving storage-only orphans. The registry now validates the ResponseBuffer middleware is in place before handling manifest pushes, enabling error recovery when artifact registration fails.

## Testing

- [x] Unit tests added to verify ResponseBuffer requirement check
- [x] Unit tests added to verify error response on artifact registration failure
- [x] Integration tests pass (existing push flow tests updated to use ResponseBuffer)
- [x] Existing tests pass (backward compatibility verified)

## Checklist

- [x] Tests updated/added


rossigee:
Closing this rather than continuing to patch it — the premise doesn't hold up under closer review, and the same critique was already raised in detail on the `goharbor/harbor` mirror of this PR (#23375) before it was redirected here.

**The guard is unreachable in production.** This route is registered as:

```go
Middleware(blob.PutManifestMiddleware()).
HandlerFunc(putManifest)
```

`blob.PutManifestMiddleware()` is built on `middleware.AfterResponse(...)`, which unconditionally wraps the writer before calling the next handler:

```go
res, ok := w.(*lib.ResponseBuffer)
if !ok {
    res = lib.NewResponseBuffer(w)
    defer res.Flush()
}
next.ServeHTTP(res, r)
```

So `putManifest` can never see a non-`*lib.ResponseBuffer` writer on this route — the `if _, isBuffer := w.(*lib.ResponseBuffer); !isBuffer { ... }` guard this PR adds never fires.

**It's also redundant even where it would matter.** `AfterResponse` already resets the buffer and sends an error instead of the buffered 201 if the hook fails:

```go
if err := hook(res, r, res.StatusCode()); err != nil {
    _ = res.Reset()
    lib_http.SendError(res, err)
}
```

That's the generic behavior every `AfterResponse`-wrapped route gets from the framework, independent of anything added in `putManifest` itself. This PR's guard and extra `buffer.Flush()` don't add protection beyond what already exists structurally upstream of the handler.

**And it doesn't address the actual issue this references (#23199 — storage-only orphans).** The blob is already committed to backend storage the moment the proxied PUT succeeds; returning a 500 instead of a 201 afterward doesn't undo that write. The orphan still exists — the fix would need to be on the GC side (detecting storage objects with no DB record), not in the response-code handling here.

The regression tests added here call `putManifest` directly with a manually-constructed `ResponseBuffer`, bypassing the real middleware chain, so they don't exercise the actual production path or demonstrate the claimed failure mode.

Happy to revisit if there's a concrete reproduction that goes through the real middleware chain (`blob.PutManifestMiddleware()` → `putManifest`) and shows a case where a client-visible 201 survives an `AfterResponse` hook failure — but as submitted, this doesn't hold up.

PR #419, opened 2026-07-05
test: Isolate CSRF plaintext HTTP env

### LINKED ISSUE #418: Fix flaky CSRF plaintext HTTP unit test in main pipeline
## Summary

The `Test` workflow on `main` is failing in `Unit Tests (DB 2/2)` for the CSRF middleware package.

Failed run: https://github.com/container-registry/harbor-next/actions/runs/28744727793

## Failure

`src/server/middleware/csrf/csrf_test.go` fails in `TestMiddlewareRejectsPlaintextHTTPOriginByDefault`:

```text
expected: 403
actual  : 200
```

## Root cause

The test is meant to verify the default path where plaintext HTTP requests are rejected unless `CSRF_PLAINTEXT_HTTP=true` is configured. However, the test does not isolate `CSRF_PLAINTEXT_HTTP` before rebuilding the CSRF middleware singleton. If the runner environment already has that variable set to `true`, the middleware marks the request with `csrf.PlaintextHTTPRequest`, the valid same-origin POST succeeds, and the test observes `200` instead of `403`.

## Fix

Explicitly set `CSRF_PLAINTEXT_HTTP=false` in the default rejection test before calling `resetMiddleware()`, so the test always exercises the default behavior regardless of the ambient runner environment.

bupd:
## Summary
- Fixes #418 by explicitly setting `CSRF_PLAINTEXT_HTTP=false` in the CSRF default rejection test
- Keeps the change test-only; no runtime behavior is changed

## Validation
- `git diff --check next/main...HEAD`
- `go test ./server/middleware/csrf`
- `go test -c -tags db ./server/middleware/csrf`

### DIFF
--- a/src/server/middleware/csrf/csrf_test.go
+++ b/src/server/middleware/csrf/csrf_test.go
@@ -79,6 +79,7 @@ func TestMiddlewareRejectsPlaintextHTTPOriginByDefault(t *testing.T) {
 		common.ExtEndpoint: "http://localhost:4500",
 	}
 	config.InitWithSettings(conf)
+	t.Setenv(csrfPlaintextHTTPEnv, "false")
 	resetMiddleware()
 	defer func() {
 		config.InitWithSettings(map[string]any{


src/server/middleware/csrf/csrf.go
```diff
@@ -95,7 +95,15 @@ func Middleware() func(handler http.Handler) http.Handler {
 }
 
 func plaintextHTTPEnabled() bool {
-	return strings.EqualFold(os.Getenv(csrfPlaintextHTTPEnv), "true")
+	if !strings.EqualFold(os.Getenv(csrfPlaintextHTTPEnv), "true") {
+		return false
+	}
+	ep, _ := config.ExtEndpoint()
+	ep = strings.ToLower(ep)
+	return strings.HasPrefix(ep, "http://localhost") ||
```
Vad1mo: `plaintextHTTPEnabled()` now does more than its name suggests. Previously it was a simple env-var check for `CSRF_PLAINTEXT_HTTP`; now it also embeds endpoint classification and dev/local policy decisions.

The function name and responsibility changed without the abstraction changing!

Vad1mo (changes requested):

src/server/middleware/csrf/csrf.go
bupd: yeah true, need to refactor this

PR #463, opened 2026-07-13
fix(portal): Make Log Rotation settings savable and surface no-op purge schedules

### LINKED ISSUE #462: Log Rotation (Clean Up): settings cannot be saved and legacy schedules silently purge nothing
## Observed problem

On **Clean Up → Log Rotation** (`/harbor/clearing-job/audit-log-purge`), with a purge schedule that predates the event-types feature (its `job_parameters` lack `include_event_types`):

- All *Event types to purge* checkboxes load unchecked and the schedule's **SAVE** button is disabled with no explanation near it — the only hint is a small subtext next to the checkboxes.
- Changing **Keep records in** (e.g. 300 → 180) is never applied: the value is only persisted through that same disabled SAVE, and outside the schedule's edit mode there is no save control at all.
- Even with valid input, re-saving an existing custom schedule fails with 400: the scheduler stores the cron with a randomized seconds field (`0 0 14 * * 0` → `55 0 14 * * 0`) which the API then rejects ("the 1st field … must be 0").
- Such legacy schedules are **silent no-ops**: with empty `include_event_types` the purge job logs "no operation selected, skip to purge audit log" and deletes nothing, while the UI shows a healthy next-run time.
- Bonus: the retention input pattern accepts negative values (`-5`), no button guard checks the valid range, and the backend only caps the max — a negative retention turns the purge into `op_time < NOW() + X hours`, i.e. deletes essentially all audit logs.

Upstream goharbor/harbor has identical logic (verified against current main), so all of this applies there too.

## What a good solution looks like

- A visible, always-available **SAVE** for retention + event types that updates the existing schedule's parameters without forcing the user through the schedule EDIT flow.
- When SAVE is disabled, say why, next to the button.
- A clear warning when an existing schedule has no event types selected ("this schedule will not purge anything").
- Normalize the cron seconds field to `0` before resubmitting stored crons so edit round-trips don't 400.
- Reject non-positive retention values in the UI guards.

Vad1mo:
Fixes #462

## Problem

On **Clean Up → Log Rotation**, settings could end up permanently unsavable: with a legacy schedule (no `include_event_types` in `job_parameters`) all event-type checkboxes load unchecked, the schedule's SAVE was silently disabled, and "Keep records in" changes were never applied because that hidden SAVE was the only persistence path. Such schedules also purge nothing at runtime while the UI shows a healthy next-run time.

### Bug fixes
- **Cron seconds round-trip** — the scheduler stores crons with a randomized seconds field (`0 0 14 * * 0` → `55 0 14 * * 0`) which the create/update API rejects, so re-saving an existing custom schedule always 400'd. Seconds are normalized to `0` before resubmission (also fixed in the shared `cron-selection` component for its remaining consumers, e.g. GC).
- **Negative retention guard** — pattern was `^[\-1-9]{1}[0-9]*$` (accepts `-5`) and the valid-range check was not wired into any button guard; the backend only caps the max, so a negative retention deletes essentially all audit logs (`op_time < NOW() + X hours`). Pattern is now positive-only and all actions validate the range.
- **No-op schedule warning** — banner when an existing schedule has zero event types selected ("This schedule will not purge anything…").

### UX rework
- Page is grouped into **Rotation Policy** (schedule, retention, event types + one SAVE), **Execution** (last completed / next scheduled + PURGE NOW / DRY RUN) and **Purge History**.
- The schedule is now an always-editable form row (type select + cron input); the EDIT → mini SAVE/CANCEL flow is gone. One SAVE validates and persists everything, creating or updating the schedule as needed (type *None* removes it).
- SAVE activates only when something actually differs from the saved state and resets after load/save.
- Status lines are stacked (`Last completed:` / `Next scheduled time:`) instead of inline.

New i18n keys (`PURGE_PARAMS_SAVED`, `NO_EVENT_TYPES_WARNING`, `ROTATION_POLICY`, `EXECUTION`) added to all 10 languages.

## Upstream

goharbor/harbor `main` has identical logic in these components (only `@if` control-flow syntax differs). The bug fixes apply there directly; the UX rework is a fork improvement that can be offered upstream separately.

## Verification

Reproduced the broken state locally (schedule row stripped of `include_event_types` in the DB) and verified in the dev portal end-to-end:

- Legacy schedule: warning banner, SAVE disabled until an event type is checked.
- Retention / event-type / schedule changes persist through the single SAVE (API confirms `audit_retention_hour`, `include_event_types`, cron/type); SAVE re-disables once pristine.
- Editing the cron in place and saving works (previously 400 on the randomized-seconds cron); invalid cron shows an inline error and blocks SAVE; negative retention blocks SAVE / PURGE NOW / DRY RUN.
- Manual purge executed successfully from the reworked page (Purge History shows SUCCESS).
- `ng test` for the touched components passes; `ng lint` clean; prettier applied.


bupd (approved):
/lgtm

PR #465, opened 2026-07-13
fix(scan): Keep SBOM accessory push on the local registry when CORE_URL has no port

Vad1mo:
`GenAccessoryArt` (`src/pkg/scan/util.go`) builds the accessory push target by string-joining the registry URL into `name.ParseReference`:

```go
name.ParseReference(fmt.Sprintf("%s/%s@%s", sq.Registry.URL, sq.Artifact.Repository, dgst))
```

go-containerregistry only treats the first path segment as a registry **iff it contains a `.` or `:`**. The SBOM handler strips the scheme from `CORE_URL` before this point (`sbom.go`), so a portless in-cluster URL (`http://harbor-core`) becomes the bare string `harbor-core` — no dot, no colon — and the whole reference parses as a **Docker Hub repository**:

```
GET https://auth.docker.io/token?scope=repository%3Aharbor-core%2F8gcr%2Ftrivy-adapter%3Apush%2Cpull&service=registry.docker.io
→ 400 Bad Request {"details":"malformed HTTP Authorization header"}
```

The robot account's credentials are sent to `auth.docker.io`, every SBOM generation fails, and the failures are disguised: they log under the `IMAGE_SCAN` job name, while vulnerability scanning keeps working (the trivy adapter re-adds the default port on its own side — that asymmetry is why only SBOM breaks).

Observed in production on three instances running the harbor-next chart (whose `harbor.core.url` helper omits the service port — flagged separately on the chart PR: container-registry/harbor-next#56 [comment](https://github.com/container-registry/harbor-next/pull/56#issuecomment-4963559423)): no `sbom.harbor` accessory created since the chart migration, every scan-all ending `Error`.

## Fix

Construct the reference from typed parts so the registry component is never re-parsed heuristically:

```go
reg, err := name.NewRegistry(registryURL, opts...)   // name.Insecure when http
accRef := reg.Repo(repository).Digest(dgst)
```

Any valid host now works with or without a port, and a dotted repository path segment can never be mistaken for a registry either. Harbor must behave correctly outside Helm too (compose, bare manifests), so this belongs in code rather than relying on installers to always embed a port — though the chart should keep the port regardless, since upstream goharbor images carry this bug in every released version.

Verified against the production repro (ggcr v0.20.7):

```
BUG portless (old code)        registry=index.docker.io  scope=repository:harbor-core/8gcr/trivy-adapter:push,pull
typed construction, portless   registry=harbor-core      scope=repository:8gcr/trivy-adapter:push,pull  scheme=http
```

## Tests

- `TestAccessoryRef`: portless single-label host, host with port, FQDN/secure scheme, dotted project name — plus the explicit `!= index.docker.io` regression assertion.
- `TestAccessoryRefInvalidRegistry`: invalid host errors.
- Pre-existing `TestGenAccessoryArt` (end-to-end against ggcr's in-memory registry) still passes unchanged.

Upstream goharbor/harbor has the identical code path in `GenAccessoryArt`; this is upstreamable as-is.

### DIFF
--- a/src/pkg/scan/util.go
+++ b/src/pkg/scan/util.go
@@ -40,6 +40,25 @@ func RemoteOptions() []remote.Option {
 	return []remote.Option{remote.WithTransport(tr)}
 }
 
+// accessoryRef builds the reference the accessory artifact is pushed to. The
+// registry host is constructed as a typed name.Registry instead of being
+// string-joined into name.ParseReference: ParseReference only treats the first
+// path segment as a registry if it contains a '.' or ':', so a portless
+// single-label host like "harbor-core" (the in-cluster core service) would be
+// parsed as a Docker Hub namespace and the push would leave the cluster,
+// sending robot credentials to auth.docker.io.
+func accessoryRef(registryURL, repository, dgst string, insecure bool) (name.Reference, error) {
+	var opts []name.Option
+	if insecure {
+		opts = append(opts, name.Insecure)
+	}
+	reg, err := name.NewRegistry(registryURL, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("parsing registry %q: %w", registryURL, err)
+	}
+	return reg.Repo(repository).Digest(dgst), nil
+}
+
 // GenAccessoryArt composes the accessory oci object and push it back to harbor core as an accessory of the scanned artifact.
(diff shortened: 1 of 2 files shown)


bupd (approved):
/lgtm

PR #481, opened 2026-07-20
fix: Locked users authenticate as nil instead of returning an error

### LINKED ISSUE #480: Locked users can authenticate as an anonymous/nil user
## Expected behavior and actual behavior

When a user account is locked due to repeated login failures, `auth.Login()` in `src/core/auth/authenticator.go` returns `(nil, nil)` instead of an error. This conflates "locked" with "not authenticated, no error" at the type level.

Both current callers of `Login()` happen to nil-check the returned user defensively today:
- `src/core/controllers/base.go` (`CommonController.Login`)
- `src/server/middleware/security/basic_auth.go` (`basicAuth.Generate`)

...but this is implicit, fragile behavior: any future or alternate caller that follows the common Go idiom of checking `err != nil` alone to decide whether authentication succeeded will authenticate a locked user with a nil `*models.User`, which will either panic downstream or, depending on what's built on top of the returned (nil-error, nil-user) pair, be treated as success.

## Steps to reproduce

1. Trigger enough failed logins for a principal to become locked (see `UserLock`/`frozenTime` in `src/core/auth/lock.go`).
2. Call `auth.Login()` for that principal.
3. Observe `(nil, nil)` returned instead of an error indicating the account is locked.

## Context

Identified while triaging `goharbor/harbor#23483`, which bundles this fix together with a much larger, unrelated `auth_mode` dispatch refactor. This issue tracks just the locked-user bug in isolation.

rossigee:
## Summary

Fixes #480.

`auth.Login()` returned `(nil, nil)` when a principal was locked due to repeated login failures, instead of an error. Callers that check `err != nil` alone (the common Go idiom) rather than explicitly nil-checking the returned user would treat a locked account as a successful, anonymous authentication.

## Change

`src/core/auth/authenticator.go`: the lock check in `Login()` now returns `NewErrAuth("user is locked due to repeated login failures")` instead of `nil, nil`, using the `ErrAuth` type that already exists in this file. Removed the `//nolint:nilnil` comment that was suppressing the linter warning on this exact pattern rather than fixing it.

Both current callers already branch on `err != nil` first:
- `src/core/controllers/base.go` (`CommonController.Login`)
- `src/server/middleware/security/basic_auth.go` (`basicAuth.Generate`)

So this is behaviorally transparent to them today - same 401/nil outcome - but now via the correct typed error instead of relying on every caller remembering to nil-check.

## Testing

Added `src/core/auth/authenticator_test.go` (`TestLoginLockedUserReturnsError`) - this path had zero existing coverage in either the file or its test file. The test registers a stub `AuthenticateHelper`, locks a principal via the package's `lock.Lock()`, calls `Login()`, and asserts an `ErrAuth` is returned with a nil user. Runs fully in-memory (`config.InitWithSettings` + the in-memory config driver), no database required.

## Type of Change
- [x] Bug fix

## Context

Extracted from `goharbor/harbor#23483`, which bundles this fix together with a much larger, unrelated `auth_mode` dispatch refactor that upstream maintainers flagged as too large to review as one PR. This is the locked-user portion in isolation.

### DIFF
--- a/src/core/auth/authenticator.go
+++ b/src/core/auth/authenticator.go
@@ -150,8 +150,7 @@ func Login(ctx context.Context, m models.AuthModel) (*models.User, error) {
 	}
 	if lock.IsLocked(m.Principal) {
 		log.Debugf("%s is locked due to login failure, login failed", m.Principal)
-		// nolint:nilnil // user is locked
-		return nil, nil
+		return nil, NewErrAuth("user is locked due to repeated login failures")
 	}
 	user, err := authenticator.Authenticate(ctx, m)
 	if err != nil {
--- a/src/core/auth/authenticator_test.go
+++ b/src/core/auth/authenticator_test.go
@@ -1,0 +1,61 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package auth
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/goharbor/harbor/src/common"
+	"github.com/goharbor/harbor/src/common/models"
+	"github.com/goharbor/harbor/src/lib/config"
+	_ "github.com/goharbor/harbor/src/pkg/config/inmemory"
+)
+
+// stubAuthenticateHelper always succeeds, so any lock rejection observed in
+// a test can only come from the lock check itself, not from Authenticate.
+type stubAuthenticateHelper struct {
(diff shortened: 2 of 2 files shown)


bupd:
Ran it:

```
cd src && go test ./core/auth/ -run TestLogin -v
--- PASS: TestLoginLockedUserReturnsError (0.00s)
ok  github.com/goharbor/harbor/src/core/auth  0.004s
```

`go build ./...` is clean. The branch is 163 commits behind `main`, but the hunk is one return statement in a function `main` hasn't touched, so the merge result should match.

Checked both callers — `src/core/controllers/base.go` and `src/server/middleware/security/basic_auth.go` — and each branches on `err != nil` before using the user. So a locked user got 401 before this change and gets 401 after it. No behavior change, which is what you want from a fix like this.

What it actually closes is the trap for the next caller: `(nil, nil)` meant that idiomatic `if err != nil { ... }` handling would fall through and either dereference nil or treat a locked account as an anonymous success. Dropping the `//nolint:nilnil` alongside it is the right call — the suppression was hiding exactly this.

Small thing: the test registers a stub into the package-global `registry` map without cleanup, so it will leak into other tests if more get added to `core/auth`. Not worth blocking on, but a `t.Cleanup` would cost nothing.

## Not tested

No end-to-end login attempt against a running Harbor with a genuinely locked account — only the unit test with an in-memory config.

Good 2.15.9 backport candidate as far as I'm concerned: one line, defensive, tested, no behavior change for existing callers.

PR #487, opened 2026-07-20
feat: Add Discord notification handler

rossigee:
## Summary

Add Discord notification handler for Harbor events.

## Release Notes

Adds new Discord webhook notification handler.

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -9047,6 +9047,10 @@ definitions:
   NotifyType:
     type: string
     description: Webhook supported notify type.
+    enum:
+      - http
+      - slack
+      - discord
     example: 'http'
   PayloadFormatType:
     type: string
--- a/src/controller/webhook/controller.go
+++ b/src/controller/webhook/controller.go
@@ -30,8 +30,8 @@ var (
 	// Ctl is a global webhook controller instance
 	Ctl = NewController()
 
-	// webhookJobVendors represents webhook(http) or slack.
-	webhookJobVendors = q.NewOrList([]any{job.WebhookJobVendorType, job.SlackJobVendorType})
+	// webhookJobVendors represents webhook(http), slack, or discord.
+	webhookJobVendors = q.NewOrList([]any{job.WebhookJobVendorType, job.SlackJobVendorType, job.DiscordJobVendorType})
 )
 
 type Controller interface {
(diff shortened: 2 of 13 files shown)


bupd:
This handler can't currently be selected by anyone — I think one line is missing.

Harbor gates the notify type at `src/server/v2.0/handler/webhook.go:488` via `isNotifyTypeSupported()`, which reads `GetSupportedNotifyTypes()`, which is built from the slice at `src/pkg/notification/notification.go:100`. On this branch that slice is unchanged:

```
$ git show <this-head>:src/pkg/notification/notification.go | sed -n '100p'
	notifyTypes := []string{notifier_model.NotifyTypeHTTP, notifier_model.NotifyTypeSlack}
```

Compare the sibling PRs:

```
#486 (AMQP):      ...NotifyTypeHTTP, NotifyTypeSlack, NotifyTypeAMQP}
#489 (Matrix):    ...NotifyTypeHTTP, NotifyTypeSlack, NotifyTypeMatrix}
#490 (Teams):     ...NotifyTypeHTTP, NotifyTypeSlack, NotifyTypeTeams}
#491 (Telegram):  ...NotifyTypeHTTP, NotifyTypeSlack, NotifyTypeTelegram}
#487 (Discord):   ...NotifyTypeHTTP, NotifyTypeSlack}            <-- missing
#488 (Email):     ...NotifyTypeHTTP, NotifyTypeSlack}            <-- missing
```

So `POST /projects/{id}/webhook/policies` with `type: "discord"` is rejected with `400 unsupported target type discord`, and the portal dropdown — which renders from `metadata?.notify_type`, sourced from the same list — never offers Discord either. Everything else (registry, topic, const, hook dispatch, job vendor, bootstrap) is wired correctly; it's just this one gate.

CI is green because nothing in the test suite goes through the create-policy API path. Worth adding a test that does, since this is the second PR in the batch with the same omission.

## Separate issue: unbounded sleep in the worker

`execute` reads `Retry-After` (or the JSON `retry_after` field) on a 429 and does `time.Sleep(time.Duration(seconds) * time.Second)` with no upper bound, inside the job worker, and `MaxCurrency()` is 1. The target address is user-supplied, so an endpoint returning `429 Retry-After: 9999999` parks a jobservice worker for about 115 days. Any project admin can do that to themselves by accident with a misconfigured endpoint. Worth a cap — `min(retryAfter, 60s)` — or just returning and letting the retry framework back off.

Also, `io.ReadAll(resp.Body)` on the error path has no `io.LimitReader`, where #490 (same shape, same author) does use `io.LimitReader(resp.Body, 1024)`.

## Batch-level note

I looked at all six handler PRs together. They pairwise conflict — each edits the same lines in nine shared files, and each adds a swagger `enum:` to `NotifyType` listing only its own type plus http/slack. The enum is the risky one: resolving those conflicts by taking one side produces a build that passes CI but silently rejects the previously-merged types at the generated-model layer. If these land serially, the enum needs to be merged by hand every time, or left off entirely (it's an open string on `main` today, with validation in `isNotifyTypeSupported`).

Given that two of six got the `notification.go:100` edit wrong, the nine-edit extension surface looks like the real problem. A base PR that collapses those nine points into one registration table would make each handler ~2 new files plus one entry, with no shared-line edits and no conflicts between them.

## Not tested

I did not run the Go tests on this branch or send a notification to a real Discord webhook — the missing registration makes the end-to-end path unreachable anyway.

PR #488, opened 2026-07-20
feat: Add Email notification handler

rossigee:
## Summary

Add Email notification handler for Harbor events.

## Release Notes

Adds new Email notification handler using SMTP.

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -9047,6 +9047,10 @@ definitions:
   NotifyType:
     type: string
     description: Webhook supported notify type.
+    enum:
+      - http
+      - slack
+      - email
     example: 'http'
   PayloadFormatType:
     type: string
--- a/src/controller/webhook/controller.go
+++ b/src/controller/webhook/controller.go
@@ -30,8 +30,8 @@ var (
 	// Ctl is a global webhook controller instance
 	Ctl = NewController()
 
-	// webhookJobVendors represents webhook(http) or slack.
-	webhookJobVendors = q.NewOrList([]any{job.WebhookJobVendorType, job.SlackJobVendorType})
+	// webhookJobVendors represents webhook(http), slack, or email.
+	webhookJobVendors = q.NewOrList([]any{job.WebhookJobVendorType, job.SlackJobVendorType, job.EmailJobVendorType})
 )
 
 type Controller interface {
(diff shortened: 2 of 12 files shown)


bupd:
This one has two independent blockers before it can work at all.

## 1. The type is never registered as supported

`isNotifyTypeSupported()` at `src/server/v2.0/handler/webhook.go:488` reads `GetSupportedNotifyTypes()`, built from the slice at `src/pkg/notification/notification.go:100`. Unchanged on this branch:

```
$ git show <this-head>:src/pkg/notification/notification.go | sed -n '100p'
	notifyTypes := []string{notifier_model.NotifyTypeHTTP, notifier_model.NotifyTypeSlack}
```

while #486/#489/#490/#491 all append their type there. So `POST .../webhook/policies` with `type: "email"` returns `400 unsupported target type email`, and the portal dropdown never shows Email. CI is green because no test exercises the create-policy API path. (#487 has the identical omission.)

## 2. The target address can never be an email address

`validateTargets` runs `utils.ParseEndpoint`, which requires an `http`/`https` scheme. This handler reads recipients from `event.Target.Address` and comma-splits them — so even with (1) fixed, any actual email address in that field is rejected at validation. `validateTargets` needs to special-case this type, which the PR doesn't do.

Worth noting the field now means two different things: a URL for five handlers, a recipient list for this one. That's worth an explicit decision rather than an overload.

## 3. Three job parameters are set and never read

`port`, `use_ssl` and `insecure_skip_verify` are passed into the job but `EmailJob.execute` only uses `subject/body/to/address/from/username/password`. So the configured SMTP port is ignored (`address` has to already be `host:port`), SSL is ignored, and the target's `skip_cert_verify` is ignored. They imply behavior the job doesn't have.

That matters more than it looks, because there is no TLS handling at all and `smtp.PlainAuth`'s `Start` refuses to authenticate over an unencrypted connection — it returns `"unencrypted connection"` unless the server advertises STARTTLS or the host is localhost. `smtp.SendMail` does opportunistic STARTTLS so it may work against servers that advertise it, but there's no way to *require* TLS and no way to use implicit-TLS on 465 at all.

## 4. Credential exposure via the job store

The SMTP username and password are read from global Harbor config and passed as **plaintext job parameters**. Job parameters are persisted in Redis and on the task row, so this moves the global SMTP password from config-only storage into the job store. That's a materially wider exposure and worth being deliberate about.

## 5. CRLF in the Subject header

The message is hand-assembled with `fmt.Sprintf("Subject: %s\r\n", subject)`. `smtp.SendMail` runs `validateLine` on the envelope from/to but not on the message body, so a CRLF in the subject injects arbitrary headers. Today `subject` is `fmt.Sprintf("Harbor Event: %s", payLoad.Type)` over a closed set of event-type constants, so it isn't exploitable — but it's one refactor away from being header injection. `net/mail`, or just stripping CRLF, closes it permanently.

## Not tested

I did not run this branch's Go tests or attempt a real SMTP send — blockers (1) and (2) make the end-to-end path unreachable. The test file that exists (`email_job_test.go`, 45 lines) covers validation only, not the send path.

Happy to re-test once the registration and `validateTargets` questions are settled.

PR #484, opened 2026-07-20
refactor(auth): remove global auth_mode switch with probe-based per-backend Match() dispatch

rossigee:
## Summary

Replace the global `AUTH_MODE` config value with per-backend detection via a new `Match(ctx)` method on each authenticator. This refactor resolves authentication lockout issues when multiple auth backends are configured and removes the need for a single global auth mode switch.

## Release Notes

- **No more lockout:** local DB admin accounts can now authenticate even when OIDC/LDAP/UAA is configured as the primary mode — `auth.Login()` iterates matching backends instead of switching on a single mode string.
- **No more stale context:** auth mode was previously pre-loaded into the request context in middleware; now each middleware or handler checks the relevant config directly.
- **Config UI simplified:** `auth_mode` and `primary_auth_mode` metadata entries are removed; the system auto-detects which backends are available.
- **Registry v2auth tightened:** Docker clients sending Basic auth are now properly redirected to the Bearer token service challenge.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Testing

- All existing auth tests pass
- Manual testing verified: local DB login works with OIDC as primary auth mode
- Unit tests added for DetectAuthMode and Match() methods

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/goharbor/harbor/blob/main/CONTRIBUTING.md).
- [x] My changes generate no new warnings or errors.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] Any dependent changes have been merged and published in downstream modules.

## Related Issues

N/A

## Screenshots (if applicable)

N/A

PR #493, opened 2026-07-20
refactor(auth): remove global auth_mode switch with probe-based per-backend Match() dispatch

### LINKED ISSUE #407: Allow database authentication when OIDC is enabled for non-OIDC users

rossigee:
Replaces PR #484, which was inadvertently closed by a branch rename (`fix/auth-dispatch-oidc-lockout-http-auth` → `refactor/auth-dispatch`, for clarity — the old name read as a narrow bugfix when this is actually an architectural refactor of the auth dispatch chain). Same commits, no content changes.

## Summary
- Removes the global `auth_mode` config switch in favor of probe-based per-backend `Match()` dispatch across LDAP/UAA/OIDC/DB/auth-proxy
- Allows database authentication for non-OIDC users when OIDC is the configured auth mode, instead of locking them out
- Fixes locked users authenticating as nil instead of returning an error
- Returns a Basic auth challenge for requests already carrying Basic auth
- Addresses CodeRabbit and cubic-dev-ai review comments from the original PR

## Test plan
- [x] Existing auth controller/config/middleware test suites updated and passing
- [x] `go build` / `go vet` clean

Closes container-registry/harbor-next#407

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -8214,6 +8214,12 @@ definitions:
         x-nullable: true
         x-omitempty: true
         description: The auth mode of current Harbor instance.
+      configured_auth_modes:
+        type: array
+        items:
+          type: string
+        x-omitempty: true
+        description: All auth backends (OIDC, LDAP, UAA, HTTP auth proxy) that currently have non-empty endpoint configuration, in dispatch priority order. auth_mode reflects only the first (active) entry; additional entries indicate leftover or conflicting configuration from other backends.
       primary_auth_mode:
         type: boolean
         x-nullable: true
--- a/src/common/http/client.go
+++ b/src/common/http/client.go
@@ -183,7 +183,7 @@ func (c *Client) GetAndIteratePagination(endpoint string, v any) error {
 	}
 
 	rv := reflect.ValueOf(v)
-	if rv.Kind() != reflect.Ptr {
+	if rv.Kind() != reflect.Ptr { // nolint:govet
 		return errors.New("v should be a pointer to a slice")
 	}
 	elemType := rv.Elem().Type()
--- a/src/common/utils/utils.go
+++ b/src/common/utils/utils.go
@@ -145,7 +145,7 @@ func ConvertMapToStruct(object any, values any) error {
 		return errors.New("nil struct is not supported")
 	}
 
-	if reflect.TypeOf(object).Kind() != reflect.Ptr {
+	if reflect.TypeOf(object).Kind() != reflect.Ptr { // nolint:govet
 		return errors.New("object should be referred by pointer")
 	}
 
--- a/src/controller/config/controller.go
+++ b/src/controller/config/controller.go
@@ -63,18 +63,21 @@ func NewController() Controller {
 	return &controller{userManager: user.Mgr}
 }
 
+// UserConfigs returns user-scoped configuration values.
 func (c *controller) UserConfigs(ctx context.Context) (map[string]*models.Value, error) {
 	mgr := config.GetCfgManager(ctx)
 	configs := mgr.GetUserCfgs(ctx)
 	return c.ConvertForGet(ctx, configs, false)
 }
 
+// AllConfigs returns all configuration values including system configs.
 func (c *controller) AllConfigs(ctx context.Context) (map[string]any, error) {
 	mgr := config.GetCfgManager(ctx)
 	configs := mgr.GetAll(ctx)
 	return configs, nil
 }
 
+// UpdateUserConfigs updates user-scoped configuration values.
 func (c *controller) UpdateUserConfigs(ctx context.Context, conf map[string]any) error {
 	if readOnlyForAll {
 		return errors.ForbiddenError(nil).WithMessage("current config is init by env variable: CONFIG_OVERWRITE_JSON, it cannot be updated")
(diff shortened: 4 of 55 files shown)


rossigee:
Addressed the outstanding items:

- **DCO**: the two newest commits were missing sign-off — rebased with `--signoff` to fix, no content changes.
- **Go Lint (gofmt)**: fixed as a side effect of the change below (the malformed block was rewritten).
- **cubic-dev-ai P1** (upgraded deployments with an explicit `auth_mode` losing that override after `common.AUTHMode` was dropped from the registered metadata list): rather than reading the legacy value straight from the driver (which would keep a permanent legacy read-path around), dropped the fallback branch entirely — it could never fire anyway once `auth_mode` stopped being writable/persisted earlier in this same refactor. Paired with a portal fix so the auth settings page derives the active backend from the new `configured_auth_modes` field instead of the now-unused `auth_mode`, and surfaces a warning when more than one backend has non-empty endpoint config (since only the highest-priority one is used for dispatch, and the others would otherwise be invisible leftover configuration from a prior setup — which is the actual operator-facing risk the P1 comment was pointing at).

bupd:
Read this one carefully because of what it deletes. Three fail-open paths, all landing on local DB password auth.

First, to be fair to the title: despite "probe-based dispatch", there is **no multi-backend probing and no ambiguity**. `loginHelpers()` switches on the single `DetectAuthMode()` result and returns exactly one helper per branch; `Match()` is only a configured-ness guard, the priority order (OIDC > LDAP > UAA > HTTP-proxy) is hard-coded and deterministic, and two backends cannot both win. An attacker can't steer which backend handles their login. That part is sound.

The problem is elsewhere.

## 1. OIDC alone falls back to DB auth

```go
case common.OIDCAuth:
    if !h.Match(ctx) {
        return []AuthenticateHelper{registry[common.DBAuth]}
    }
```

LDAP, UAA and HTTPAuth all return an **error** in the equivalent situation. OIDC returns the DB helper. So an admin who blanks or mistypes the OIDC endpoint on an SSO-only instance silently converts it into a local-password instance, and every stored password hash becomes a valid credential. That should fail closed like its siblings.

## 2. A config read error is indistinguishable from "not configured"

`ConfiguredAuthBackends()` `log.Debugf`s and omits the backend when the config load **errors**, exactly as it does when the value is genuinely empty. A transient config-manager or DB blip during startup therefore yields `DetectAuthMode() == db_auth` for the whole process. Same outcome as (1), but triggered by infrastructure rather than by an admin.

## 3. Empty `RobotPrefix` routes every login to DB auth

```go
if strings.HasPrefix(m.Principal, config.RobotPrefix(ctx)) { ... DB auth ... }
```

`config.RobotPrefix()` is `DefaultMgr().Get(ctx, common.RobotNamePrefix).GetString()`. If that returns `""` — which it will if the config manager isn't ready, see (2) — then `strings.HasPrefix(anything, "")` is `true` and **every** login on the instance goes to DB auth, bypassing LDAP/OIDC/UAA entirely. This is the sharpest of the three because it needs no admin mistake at all.

## Also

- `registry[common.DBAuth]` is dereferenced without a nil check in both `authenticateWithLock` and the `default:` branch of `loginHelpers`.
- Deleting `common.AUTHMode` and `common.PrimaryAuthMode` from `ConfigList` in `src/lib/config/metadata/metadatalist.go` removes persisted config keys that operators set via `AUTH_MODE`. That's a breaking config change on top of a refactor, and `auth_mode` staying in swagger as read-only while `configured_auth_modes` appears alongside it makes the contract ambiguous for API consumers.
- **Zero new test functions.** `grep '^+func Test'` over the diff returns nothing — only existing suites were modified. None of the three paths above is covered.

## Suggestion

Split or close. This PR also re-implements #481 and #497 verbatim inside itself, so it's a superset that will conflict with both. Those two are small, tested and independently reviewable — land them on their own. Then resubmit the `Match()` work as its own PR once OIDC fails closed, config errors are distinguished from unconfigured, and the empty-`RobotPrefix` case is guarded. The dispatch idea is reasonable; it's the failure modes that need to change, and they're much easier to review without 55 files around them.

Not a 2.15.9 candidate: it's a refactor by its own title and it deletes persisted config keys.

## Not tested

I did not build or run this branch — it's CONFLICTING against `main`, and the three findings above are static reads of the control flow rather than reproductions. Worth confirming (3) with a test that sets `RobotNamePrefix` to `""` before this goes further, since it's the cheapest to prove either way.

PR #496, opened 2026-07-21
fix(tls): Honor Mounted Custom CA Certificates in Scratch-Based Images

rossigee:
## Summary
- `core`, `jobservice`, `registryctl`, and `exporter` are built `FROM scratch` with no shell/`update-ca-certificates`, so the long-standing `/harbor_cust_cert` mount convention (used by the Helm chart's `caBundleSecretName`) was silently inert — nothing in these images ever read that directory.
- Surfaced as a 500 on the artifact Dockerfile/Build History tabs: `core`'s outbound TLS connection to the registry's token realm failed with `certificate signed by unknown authority` whenever that realm's certificate was issued by a private CA.
- Adds `LoadCustomCACertificates()`, called first thing in each service's `main()`, which merges any `.crt`/`.pem` files found in the mount directory into the baked-in system CA bundle and points `SSL_CERT_FILE` at the merged result — `crypto/x509` reads that env var automatically, so no other application code needs to change. It's a no-op when the mount is absent or empty (the common case).
- The mount directory defaults to `/harbor_cust_cert` (unchanged, for Helm chart compatibility) but can be overridden via the new `CUSTOM_CA_CERT_DIR` env var.
- The merged bundle is written to a new directory (`/etc/ssl/harbor-custom-ca`) created empty and owned by the non-root `harbor` user at build time, kept separate from the baked-in `/etc/ssl/certs/ca-certificates.crt` (which stays root-owned and read-only) — avoids giving the running process write access to its own trusted root store.
- Adds `docs/custom-ca-certificates.md` covering the mount convention, the env var override, and the `harbor-registry` exception (its binary is unmodified upstream `distribution` source, so it needs `SSL_CERT_FILE` set directly rather than going through this mechanism).

## Release Notes

Fixes custom CA certificates mounted at `/harbor_cust_cert` (or `CUSTOM_CA_CERT_DIR` if overridden) being silently ignored by `core`, `jobservice`, `registryctl`, and `exporter`, which caused internal TLS connections between these services to fail with `certificate signed by unknown authority` for deployments using a private CA. See [Custom CA Certificates](docs/custom-ca-certificates.md) for configuration details.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./common/http/...` clean
- [x] `golangci-lint run ./common/http/...` clean (pre-existing, unrelated `govet` finding in `client.go` not touched by this PR)
- [x] `hadolint` clean on all 4 changed Dockerfiles
- [x] `go test ./common/http/...` passes; new code coverage: `LoadCustomCACertificates` 100%, `resolveCustomCACertDir` 100%, `loadCustomCACertificates` 96.8%
- [x] Verified live: fixed a real production 500 on the Dockerfile/Build History artifact tabs caused by this exact gap

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ We're developing Harbor Next as a [community proposal](https://github.com/goharb
 - Continuous delivery
 - Easy Contributor onboarding with out of the box dev environments
 - Multi-architecture artifacts
-- Scratch images with minimal size and attack surface.
+- Scratch images with minimal size and attack surface — see [Custom CA Certificates](docs/custom-ca-certificates.md) for configuring internal TLS trust in these images.
 - Use of Docker Distribution V3
 - Replicate images to SFTP endpoints
 - Harbor Satellite Support
--- a/dockerfile/core.dockerfile
+++ b/dockerfile/core.dockerfile
@@ -7,7 +7,7 @@ RUN addgroup -S -g 10000 harbor && adduser -S -G harbor -u 10000 harbor && \
 FROM scratch
 COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=certs /etc/passwd /etc/group /etc/
-COPY --from=certs /harbor-ca-writable /etc/ssl/harbor-custom-ca
+COPY --chown=10000:10000 --from=certs /harbor-ca-writable /etc/ssl/harbor-custom-ca
 ARG TARGETARCH
 COPY bin/linux-${TARGETARCH}/lprobe /lprobe
 COPY bin/linux-${TARGETARCH}/core /core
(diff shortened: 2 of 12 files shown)


Vad1mo:
Heads up while this is still open: the loader covers `/harbor_cust_cert`, but the goharbor chart's `internalTLS` mounts the CA at `/etc/harbor/ssl/<component>/ca.crt` and announces it via `INTERNAL_TLS_TRUST_CA_PATH`, which neither this PR nor the current code reads, so internal TLS stays broken after merge. Details and a repro in #913. Extending `LoadCustomCACertificates()` to also take `INTERNAL_TLS_TRUST_CA_PATH` (and giving registry an `SSL_CERT_DIR`) would let one fix close both.

PR #497, opened 2026-07-22
fix(auth): return Basic challenge for requests already carrying Basic auth

rossigee:
Fixes the v2auth middleware to return a Basic challenge when the request already carries a Basic Authorization header.

Currently, getChallenge() only special-cases this for the catalog path (/v2/_catalog). For other paths, it always builds a Bearer challenge pointing at the token service — but a client sending Basic auth isn't following the OCI/Docker Bearer token flow, so pointing it at the token service is incorrect.

This change moves the check up as an unconditional early return, matching the existing catalog short-circuit.

## Release Notes

- Fixed: v2auth now returns Basic challenge for requests already carrying Basic auth

### DIFF
--- a/src/server/middleware/v2auth/auth.go
+++ b/src/server/middleware/v2auth/auth.go
@@ -96,6 +96,14 @@ func getChallenge(req *http.Request, accessList []access) string {
 		// Return basic auth challenge by default, incl. request to '/v2/_catalog'
 		return `Basic realm="harbor"`
 	}
+
+	// A request that already carries Basic credentials isn't following the
+	// OCI/Docker Bearer token flow, so challenge it with Basic too instead of
+	// pointing it at the token service.
+	if strings.HasPrefix(auth, "Basic ") {
+		return `Basic realm="harbor"`
+	}
+
 	// No auth header, treat it as CLI and redirect to token service
 	tokenSvc, err := tokenSvcURL(req)
 	if err != nil {
--- a/src/server/middleware/v2auth/auth.go
+++ b/src/server/middleware/v2auth/auth.go
@@ -92,19 +92,16 @@ func (rc *reqChecker) projectID(ctx context.Context, name string) (int64, error)
 func getChallenge(req *http.Request, accessList []access) string {
 	logger := log.G(req.Context())
 	auth := req.Header.Get(authHeader)
-	if len(auth) > 0 || lib.V2CatalogURLRe.MatchString(req.URL.Path) {
-		// Return basic auth challenge by default, incl. request to '/v2/_catalog'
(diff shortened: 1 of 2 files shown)


rossigee:
Addressed the coderabbitai/cubic-dev-ai findings: the added Basic-challenge branch was unreachable dead code because the preceding `len(auth) > 0` check already caught every non-empty `Authorization` header, including Bearer — so Bearer-authenticated requests needing a token refresh were incorrectly challenged with Basic, and the test suite asserted that wrong behavior instead of catching it.

Folded the Basic-header check and the `/v2/_catalog` check into a single guard so Bearer (or no header) on any other path correctly falls through to the token-service challenge, and added a regression test for the previously-untested Bearer-on-non-catalog-path case.

bupd:
Ran the tests here. First attempt looked like a failure and wasn't — noting it because it will catch the next person too: `src/server/middleware/v2auth/auth_test.go` is `//go:build db`, so plain `go test ./server/middleware/v2auth/` reports `[no test files]`, and with `-tags db` but no `POSTGRES_MIGRATION_SCRIPTS_PATH` it dies with `failed to open source, "file://migrations/postgresql/"`.

With both set:

```
cd src && go test -tags db -count=1 -p 1 -v ./server/middleware/v2auth/ -run 'TestGetChallenge|TestMiddleware'
ok  github.com/goharbor/harbor/src/server/middleware/v2auth  0.113s
```

All nine `TestGetChallenge` subtests pass, including the new one:

```
--- PASS: TestGetChallenge/Request_to_'/v2'_carrying_a_Bearer_auth_header_should_still_get_a_token_service_challenge,_not_Basic
```

`go build ./...` is clean at this head (163 commits behind `main`, but `strings` is already imported on `main` so the hunk still applies).

On the "does this downgrade anything" question — it does the opposite. The old `len(auth) > 0` answered `Basic realm="harbor"` to *any* request carrying *any* Authorization header, including a Bearer-carrying OCI client whose token had expired or lacked scope. That is itself a nudge toward Basic. Narrowing to `strings.HasPrefix(auth, "Basic ")` sends Bearer clients to the token service, which is the correct OCI flow, and only answers Basic to clients that already chose it. Net effect is strictly fewer Basic challenges. Non-Basic/non-Bearer schemes (`Negotiate`, say) now get a Bearer challenge instead of Basic, which is harmless.

## Not tested

No real `docker login` / `docker pull` / `podman push` against a running Harbor. The unit test asserts the header string; it does not prove a real client follows the Bearer challenge correctly. Since this changes challenge negotiation for every OCI client, that smoke test is worth doing before it rides into a patch release.

One coordination note: the identical hunk is duplicated inside #493. Land one, not both.

PR #499, opened 2026-07-22
fix(jobservice): handle nil config manager instead of segfault

rossigee:
## Summary

- `DefaultMgr()` could return `nil` when the REST config manager fails to register (missing `CORE_URL` or `JOBSERVICE_SECRET` env vars).
- Both the config manager and the jobservice startup now handle this gracefully with a clear error message instead of a nil pointer dereference.

## Release Notes

Fix a jobservice segfault on startup when `CORE_URL` or `JOBSERVICE_SECRET` are missing/misconfigured; it now fails with a clear panic message instead of a nil pointer dereference.

## Checklist

### DIFF
--- a/src/jobservice/main.go
+++ b/src/jobservice/main.go
@@ -50,8 +50,12 @@ func main() {
 	lib.StartPprof()
 
 	cfgLib.DefaultCfgManager = common.RestCfgManager
+	mgr := cfgLib.DefaultMgr()
+	if mgr == nil {
+		panic("failed to initialize config manager - check CORE_URL and JOBSERVICE_SECRET env vars")
+	}
 	if err := retry.Retry(func() error {
-		return cfgLib.DefaultMgr().Load(context.Background())
+		return mgr.Load(context.Background())
 	},
 		retry.InitialInterval(500*time.Millisecond),
 		retry.MaxInterval(10*time.Second),
--- a/src/lib/config/config.go
+++ b/src/lib/config/config.go
@@ -81,6 +81,9 @@ func DefaultMgr() Manager {
 	if err != nil {
 		log.Error("failed to get config manager")
 	}
+	if manager == nil {
+		log.Error("config manager is nil")
+	}
 	return manager
 }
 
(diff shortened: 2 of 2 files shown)


rossigee:
Addressed: `config.Load()` and `config.Upload()` in `src/lib/config/config.go` now check for a nil manager and return an error instead of calling `DefaultMgr().Load()`/`.UpdateConfig()` unconditionally. All existing callers of `Load()` (`core/main.go`, `jobservice/logger/sweeper_controller.go`, `pkg/registry/interceptor/readonly/interceptor.go`) already check the returned error, so this closes off the other nil-pointer paths with no caller changes needed. Left the much larger set of unguarded `DefaultMgr().Get(...)` call sites across `systemconfig.go`/`userconfig.go` out of scope — that's pre-existing, unrelated to this PR's diff, and would need its own pass.

rossigee:
Filed container-registry/harbor-next#501 to track the remaining ~40 unguarded `DefaultMgr().Get(...)` call sites in `systemconfig.go`/`userconfig.go` as a separate, larger design question rather than folding it into this PR.

bupd:
Built and ran this:

```
cd src && go build ./...          # clean
cd src && go test ./lib/config/...
?   github.com/goharbor/harbor/src/lib/config          [no test files]
ok  github.com/goharbor/harbor/src/lib/config/metadata  0.003s
?   github.com/goharbor/harbor/src/lib/config/models    [no test files]
```

No regressions. Worth being explicit that **nothing here is actually tested** — `lib/config` has no test file, `./jobservice` has no test for `main()`, and the PR adds none. The change is small and obviously correct by inspection, but "CI is green" means "nothing broke", not "the nil path works".

The behaviour change is good: a missing `CORE_URL` or `JOBSERVICE_SECRET` currently produces a nil-pointer dereference with a stack trace pointing somewhere unhelpful, and after this it produces a message naming the actual problem.

Two small things, neither worth blocking on:

- `panic()` in `main()` where `log.Fatalf` would be the house style and would print without the goroutine dump. It is genuinely unrecoverable at that point either way.
- `Upload()` is documented as "used by testing", so the nil guard there is dead weight.

## Not tested

The nil-config-manager path itself — I didn't start jobservice with `CORE_URL` unset to confirm the new message actually surfaces. Cheap to add as a test on `Load()`/`Upload()` returning the error, which would at least pin the `lib/config` half.

Reads like a reasonable 2.15.9 backport to me: tiny, defensive, turns a crash into something diagnosable.

PR #500, opened 2026-07-22
fix: handle wildcard permissions in robot token authorization

rossigee:
## Summary

- When robots/PATs are created with `/project/*` scope (all projects), the permission system was unable to match them against specific project namespaces.
- The namespace parser expected a numeric project ID but received a wildcard (`*`), causing the permission check to always fail with "denied: requested access to the resource is denied".
- Adds wildcard pattern detection in `ResourceAllowedInNamespace` to match `/project/*/*` patterns against any project namespace by comparing namespace kinds instead of requiring exact identity matches.

## Related upstream reports

This class of bug has been reported upstream against goharbor/harbor without a code-level fix landing:
- goharbor/harbor#21922 — "System robot account denied creating project robot account" (closed not-planned)
- goharbor/harbor#21406 — "Error when creating project Robot Accounts using System Robot Accounts" (closed, unresolved at the code level)

Both describe a system-level robot/PAT with all-projects scope being denied against a specific project — the same wildcard-vs-namespace matching gap fixed here.

## Release Notes

Fix robot accounts/PATs created with "all projects" (`/project/*`) scope being unable to push images due to a namespace-matching bug that only handled numeric project IDs.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ make/dev/adminserver/harbor_adminserver
 make/dev/core/harbor_core
 make/dev/jobservice/harbor_jobservice
 make/photon/*/binary/
+make/photon/*/harbor_*
 make/photon/prepare/versions
 
 src/adminserver/adminserver
--- a/src/pkg/permission/types/namespace.go
+++ b/src/pkg/permission/types/namespace.go
@@ -15,6 +15,7 @@
 package types
 
 import (
+	"strings"
 	"sync"
 )
 
@@ -67,7 +68,20 @@ func NamespaceFromResource(resource Resource) (Namespace, bool) {
 }
 
 // ResourceAllowedInNamespace returns true when resource's namespace equal the ns
+// It also handles wildcard resources like /project/*/repository which should match any project namespace
 func ResourceAllowedInNamespace(resource Resource, ns Namespace) bool {
+	resourceStr := resource.String()
+	// Check for wildcard pattern like /project/*/repository
+	if strings.Contains(resourceStr, "/*/") {
+		// Extract the kind from the resource path (e.g., "project" from "/project/*/repository")
+		parts := strings.Split(resourceStr, "/")
+		if len(parts) >= 2 {
+			kind := parts[1]
+			return kind == ns.Kind()
+		}
+		return false
+	}
+
 	n, ok := NamespaceFromResource(resource)
 	if ok {
 		return n.Kind() == ns.Kind() && n.Identity() == ns.Identity()
--- a/src/pkg/permission/types/namespace_test.go
+++ b/src/pkg/permission/types/namespace_test.go
@@ -1,0 +1,86 @@
+// Copyright Project Harbor Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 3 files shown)


bupd:
I tested this and the test passes, but I think it is testing the wrong thing — the code path it changes is not the one the reported bug goes through.

## What I ran

```
cd src && go test ./pkg/permission/types/ -v
--- PASS: TestResourceAllowedInNamespace_Wildcard (0.00s)
ok  github.com/goharbor/harbor/src/pkg/permission/types  0.002s
```

`go build ./...` is clean. So the change compiles and its own test passes.

## System robots never reach this function

`ResourceAllowedInNamespace` has exactly one non-test caller: `filterRobotPolicies` in `src/common/security/robot/context.go`. That is passed as a filter only in the `else` branch:

```go
if s.robot.Level == robot.LEVELSYSTEM {
    ...
    evaluators = evaluators.Add(rbac_project.NewEvaluator(s.ctl,
        rbac_project.NewBuilderForPolicies(s.GetUsername(), proPolicies)))   // no filter
    ...
} else {
    s.evaluator = rbac_project.NewEvaluator(s.ctl,
        rbac_project.NewBuilderForPolicies(s.GetUsername(), accesses, filterRobotPolicies))
}
```

A system-level robot takes the first branch and never calls `filterRobotPolicies`, so it never reaches `ResourceAllowedInNamespace` at all. If the symptom you're chasing is a system robot with `/project/*` being denied against a specific project, the cause is somewhere in that first branch, not here.

## What the change does do

On the project-robot path, the new pre-check returns `true` for **every** project namespace whose kind matches, because it never compares the namespace identity:

```go
if strings.Contains(resource.String(), "/*/") {
    parts := strings.Split(resource.String(), "/")
    return parts[1] == ns.Kind()
}
```

`strings.Contains(s, "/*/")` is also positionally blind — it matches a `*` anywhere in the path, not just in the project-ID slot. So any policy resource containing `/*/` gets admitted into every project's policy set.

I rate this moderate rather than confirmed-exploitable: `toScope()` in `src/controller/robot/controller.go` will happily produce `/project/*` for `Kind: LEVELPROJECT, Namespace: "*"`, and `validate()` in `src/server/v2.0/handler/robot.go` does not reject `"*"` for a project-level robot, but `requireAccess` → `RequireProjectAccess(ctx, "*", …)` should fail to resolve a project literally named `*` at creation time. Still, that is one code path or one hand-written DB row away from being cross-project push/pull escalation, in exchange for fixing nothing.

If the wildcard case genuinely needs handling here, I'd anchor on the namespace segment — parse the resource and special-case an `*` *identity* against `ns.Identity()` — rather than substring-matching the whole path. And the test needs the negative case: `/project/1/repository` must not match project 2.

## Minor

The `.gitignore` hunk adds `make/photon/*/harbor_*`, but this fork has no `make/photon/` tree — it builds from `dockerfile/`. Looks like it came along from upstream; worth dropping.

## Not tested

I did not stand up a Harbor and try to reproduce the original symptom with a system robot. That is the thing worth doing next — the fix should be re-derived from a reproduction against `src/common/security/robot/context.go`, not from this function.

PR #502, opened 2026-07-22
fix(token): use RFC 7638 JWK thumbprint for key IDs, not libtrust's format

rossigee:
## Summary

`generateKeyID` (added for PKCS8/ECDSA key support without a libtrust dependency) computed a key ID as `SHA256(DER pubkey)[:8]` in plain hex. An unmodified docker `distribution` registry computes its own trusted key ID from `rootcertbundle` using RFC 7638 JSON Web Key Thumbprint — SHA256 of a minimal canonical JSON encoding of the key's public parameters, base64url-encoded — a completely different format.

Since the two never matched for any key, every token this service issued was silently rejected by the registry with `token signed by untrusted key with ID: ...`, even though the underlying key material was identical.

## Root cause

As of `distribution` v3.1.1 (this project's `DISTRIBUTION_VERSION`), `rootCerts` are matched by `GetJWKThumbprint`, not libtrust's SHA256-truncated, colon-grouped base32 fingerprint (which `distribution` moved off of entirely). Confirmed directly against `distribution/distribution`'s v3.1.1 `registry/auth/token` source and reproduced its `GetJWKThumbprint` computation against a live deployment's actual tokens.

## Testing

Tests cross-validate `generateKeyID`'s output against `distribution/distribution/v3`'s real `GetJWKThumbprint` directly as ground truth, for RSA, ECDSA and Ed25519 keys, so this class of regression can't recur silently.

## Release Notes

Fix registry tokens being silently rejected with "token signed by untrusted key with ID: ..." due to a key-ID format mismatch with the `distribution` registry (RFC 7638 JWK thumbprint vs. libtrust's format).

## Checklist


rossigee:
Confirmed and fixed — thanks for catching this. `signingMethod` was accidentally dropped as an unused constant during the refactor, and the call site silently passed `""` instead, which would have broken all registry token issuance (`unknown sign method: `). Restored the constant and the call site.

PR #534, opened 2026-07-31
feat(artifact): Recognize standard SPDX and CycloneDX OCI SBOM types

### LINKED ISSUE #507: feat(artifact): Recognize standard SPDX and CycloneDX OCI SBOM types
## Problem

Harbor's SBOM processor is registered only for the Harbor-specific type:

`application/vnd.goharbor.harbor.sbom.v1`

Standards-based SBOM referrers pushed with ORAS use the document media type as the OCI 1.1 `artifactType`. Harbor records these as generic `subject.accessory` artifacts instead of classifying them as SBOMs.

| Format | OCI `artifactType` | Layer `mediaType` | Config `mediaType` |
|---|---|---|---|
| SPDX JSON | `application/spdx+json` | `application/spdx+json` | `application/vnd.oci.empty.v1+json` |
| CycloneDX JSON | `application/vnd.cyclonedx+json` | `application/vnd.cyclonedx+json` | `application/vnd.oci.empty.v1+json` |

```bash
oras attach \
  --artifact-type application/spdx+json \
  registry.example.com/project/image@sha256:DIGEST \
  sbom.spdx.json:application/spdx+json

oras attach \
  --artifact-type application/vnd.cyclonedx+json \
  registry.example.com/project/image@sha256:DIGEST \
  bom.cdx.json:application/vnd.cyclonedx+json
```

## Expected

- Register both standard media types as SBOM artifact types.
- Classify matching OCI referrers as `SBOM` in the API and UI.
- Expose the single SBOM layer through the existing SBOM addition path.
- Retain `application/vnd.goharbor.harbor.sbom.v1` for compatibility.
- Cover both OCI 1.1 `artifactType` and the legacy fallback where artifact type is derived from `config.mediaType`.

This issue is classification only. Externally pushed SBOMs must not automatically become trusted scan inputs. Provenance and scan reuse remain separate from #461.

Akarshkushwaha:
## Summary

This adds support for `application/spdx+json` and `application/vnd.cyclonedx+json` artifact types, allowing Harbor to properly classify industry-standard SBOMs pushed via standard OCI registries (e.g. using `oras attach`) while retaining full backwards compatibility with Harbor's legacy SBOM format.

## Related Issues

Fixes #507

## Type of Change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Documentation (`docs:`)
- [ ] Tests (`test:`)

## Release Notes

Harbor now natively recognizes standard SPDX (`application/spdx+json`) and CycloneDX (`application/vnd.cyclonedx+json`) OCI SBOM formats. When industry-standard SBOMs are pushed directly to Harbor via standard OCI clients, they will now be properly classified and displayed as SBOM accessories in the UI and API.

## Testing

## Checklist

### DIFF
--- a/src/controller/artifact/processor/sbom/sbom.go
+++ b/src/controller/artifact/processor/sbom/sbom.go
@@ -33,14 +33,25 @@ const (
 	ArtifactTypeSBOM = "SBOM"
 	// processorMediaType is the media type for SBOM, it's scope is only used to register the processor
 	processorMediaType = "application/vnd.goharbor.harbor.sbom.v1"
+	// MediaTypeSPDX is the SPDX SBOM media type
+	MediaTypeSPDX = "application/spdx+json"
+	// MediaTypeCycloneDX is the CycloneDX SBOM media type
+	MediaTypeCycloneDX = "application/vnd.cyclonedx+json"
 )
+
+var SupportedMediaTypes = []string{
+	processorMediaType,
+	MediaTypeSPDX,
+	MediaTypeCycloneDX,
+}
 
 func init() {
 	pc := &Processor{}
 	pc.ManifestProcessor = base.NewManifestProcessor()
-	if err := processor.Register(pc, processorMediaType); err != nil {
-		log.Errorf("failed to register processor for media type %s: %v", processorMediaType, err)
-		return
+	for _, mt := range SupportedMediaTypes {
+		if err := processor.Register(pc, mt); err != nil {
+			log.Errorf("failed to register processor for media type %s: %v", mt, err)
+		}
 	}
 }
 
@@ -77,9 +88,17 @@ func (m *Processor) AbstractAddition(_ context.Context, art *artifact.Artifact, 
 	if err != nil {
 		return nil, err
 	}
+	contentType := art.ResolveArtifactType()
+	switch contentType {
+	case MediaTypeSPDX, MediaTypeCycloneDX:
+		// use dynamically resolved type
+	default:
+		contentType = processorMediaType
+	}
+
 	return &processor.Addition{
 		Content:     content,
-		ContentType: processorMediaType,
+		ContentType: contentType,
 	}, nil
 }
 
--- a/src/server/middleware/subject/subject.go
+++ b/src/server/middleware/subject/subject.go
@@ -51,6 +51,10 @@ var (
 
 	// media type of harbor sbom
 	mediaTypeHarborSBOM = "application/vnd.goharbor.harbor.sbom.v1"
+	// media type of spdx sbom
+	mediaTypeSPDX = "application/spdx+json"
+	// media type of cyclonedx sbom
+	mediaTypeCycloneDX = "application/vnd.cyclonedx+json"
 
 	// source of accessory artifact is local, means the accessory is created by harbor itself
 	sourceLocal = "local"
@@ -166,7 +170,7 @@ func Middleware() func(http.Handler) http.Handler {
 				accData.Type = model.TypeNotationSignature
 			case mediaTypeCosignConfig, mediaTypeCosignArtifactType:
 				accData.Type = model.TypeCosignSignature
-			case mediaTypeHarborSBOM:
+			case mediaTypeHarborSBOM, mediaTypeSPDX, mediaTypeCycloneDX:
 				accData.Type = model.TypeHarborSBOM
 			}
 			if subjectArt != nil {


bupd:
Thanks for the PR @Akarshkushwaha — I verified it against a live Harbor and the classification part works exactly as described: SPDX and CycloneDX referrers (both via OCI 1.1 `artifactType` and the `config.mediaType` fallback) now show up as SBOM accessories, the `/additions/sbom` endpoint serves them with the right `Content-Type`, and the legacy `application/vnd.goharbor.harbor.sbom.v1` path keeps working. Nice work.

However, there is one blocking issue that needs to be addressed before we can merge:

### Blocking: reusing `model.TypeHarborSBOM` makes SBOM generation delete externally attached SBOMs

`deleteSBOMAccessory` (`src/pkg/scan/sbom/sbom.go`) deletes **every** accessory of type `sbom.harbor` before Harbor generates its own SBOM. Since this PR classifies external SPDX/CycloneDX referrers as `sbom.harbor` too, triggering "Generate SBOM" now silently destroys user-attached SBOMs.

Verified repro on this branch:

1. Push an image, then `oras attach --artifact-type application/spdx+json ...` an external SPDX SBOM → accessory created with type `sbom.harbor`.
2. `POST /api/v2.0/projects/{p}/repositories/{r}/artifacts/{digest}/scan` with `{"scan_type":"sbom"}` (Trivy).
3. The external SBOM accessory is deleted mid-scan; a `GET` on its artifact returns 404. Only Harbor's generated SBOM remains.

This couldn't happen before this PR (external SBOMs were `subject.accessory` and untouched by that cleanup). #507 also explicitly called out that externally pushed SBOMs must stay separate from Harbor's own scan pipeline.

**Requested change:** introduce a distinct accessory type for externally pushed SBOMs (e.g. `sbom.external` / `TypeExternalSBOM`) and map `application/spdx+json` / `application/vnd.cyclonedx+json` to that in the subject middleware — or alternatively filter `deleteSBOMAccessory` so it only removes Harbor-generated SBOMs. The distinct type is the cleaner option and keeps the UI able to tell the two apart later.

### Also needed

- **Tests:** please add unit coverage for the new media types — the processor registration/`AbstractAddition` content-type switch in `src/controller/artifact/processor/sbom`, and the new middleware classification cases (`src/server/middleware/subject` currently has no test file at all). Please include a regression test covering the deletion scenario above.

### Minor nits

- The SPDX/CycloneDX media-type constants are re-declared in `subject.go`; please reference the constants from the sbom processor package instead of duplicating them.
- `SupportedMediaTypes` is exported but only used within the package — either use it from the middleware or keep it unexported.

Happy to re-review once the accessory type question is sorted — the core of the change is solid.

bupd:
There's a data-loss path here that I think blocks this in its current form.

`src/pkg/scan/sbom/sbom.go:255-274` (`deleteSBOMAccessory`) walks a subject's accessories and calls `artifactCtl.Delete` on **every** accessory typed `TypeHarborSBOM`. This PR maps `application/spdx+json` and `application/vnd.cyclonedx+json` to `model.TypeHarborSBOM` in `src/server/middleware/subject/subject.go`. So once a user pushes their own SPDX or CycloneDX SBOM with `oras attach`, the next Harbor-initiated SBOM generation on that subject **deletes the user's artifact**.

The accessory model already carries a `source` (`sourceLocal` / `sourceProxyCache`) that isn't consulted there. Either a distinct accessory type for externally-pushed SBOMs, or a source check in `deleteSBOMAccessory`, would close it — but reusing `TypeHarborSBOM` for third-party content means Harbor can't tell its own artifacts from the user's.

## Two smaller things

**Incomplete media-type registration.** `src/controller/artifact/controller.go:69-96` defines `scannableContentTypes`, and under its `// SBOMs` comment it already lists `application/vnd.cyclonedx+json` but **not** `application/spdx+json`. This PR adds first-class SPDX support without closing that gap, so the two formats take different paths through `isUnscannable`.

**The registration loop.** `processor.Register` is already variadic (`Register(processor, mediaTypes...)`), so the new loop is redundant. More to the point, swapping `return` for `continue` on error means a duplicate-registration conflict is now only logged — a real misconfiguration becomes invisible at startup. Also `SupportedMediaTypes` is an exported, mutable package-level slice with no consumer outside `init()`.

## Tests

The PR description ticks "Unit tests added/updated", but no test file is changed. Both touched packages already have `sbom_test.go` and `subject_test.go`, so there's no infrastructure reason not to. The suggested command is:

```
go test ./src/controller/artifact/processor/sbom/... ./src/server/middleware/subject/...
```

A test that pushes an SPDX accessory and then triggers SBOM generation would catch the deletion issue directly.

## Not tested

I did not run this branch — I'd rather the accessory-type question be settled first, since it changes what the tests should assert.

Not a 2.15.9 candidate: it's `feat:` and it changes how existing artifacts are classified, which is not patch-release material.

Issue #44, opened 2026-03-05
Auto-generate artifacthub.io/images annotation in Chart.yaml


Vad1mo:
## Summary

ArtifactHub supports an `artifacthub.io/images` annotation in `Chart.yaml` that lists container images used by the chart. This enables ArtifactHub to run vulnerability scanning (Trivy) on listed images and display results on the package page.

Currently this annotation is not present because maintaining it manually drifts out of sync with actual image references in `values.yaml`.

## Proposal

Add a CI step (or Taskfile target) that:

1. Reads `appVersion` from `Chart.yaml`
2. Extracts image repository names from `values.yaml` (or a defined list of components)
3. Generates the `artifacthub.io/images` annotation block with correct tags
4. Updates `Chart.yaml` in-place before chart packaging/publishing

This keeps the annotation accurate without manual maintenance.

## References

- [ArtifactHub Helm annotations docs](https://artifacthub.io/docs/topics/annotations/helm/)
- Chart location: `deploy/chart/Chart.yaml`

PR #42, opened 2026-03-05
feat(compose): production compose, images, devenv overhaul with DHI base images, non-root containers, production compose

Vad1mo:
Portal image overhaul — Migrated portal to Docker Hardened Images (DHI) nginx base. Dropped the standalone nginx.dockerfile and merged its role into portal.dockerfile. Portal now serves both static SPA files and
reverse-proxies to Core (in Compose), replacing the need for a separate nginx container.

Docker Compose deployment — Added deploy/compose/ as a production-ready Compose deployment with TLS enabled by default, .env.example, local example variant (example/local/ with TLS disabled), and documentation.
Removed the old devenv/docker-compose.release.yml overlay.

Config consolidation — Moved service configs (jobservice, registry, registryctl, nginx proxy/tls) from deploy/compose/config/ and devenv/ into a single config/ directory at the repo root. Dev and prod share the
same files; dev overrides via env vars only.

Dockerfile hardening — All images run as non-root (CIS-DI-0001). Scratch images create a harbor user (UID 10000). Added sentinel defaults to ARGs used in FROM. Fixed token key file permissions (chmod 644 for
cross-UID access).

Dev environment cleanup — Removed release overlay and nginx release config from devenv. Updated air configs, port mappings, and debug port docs. Tidied .gitignore for deploy secrets.

## Test Plan

- [x] `docker compose -f deploy/compose/docker-compose.yaml --env-file deploy/compose/.env.example config --quiet` validates
- [x] `docker compose -f deploy/compose/docker-compose.yaml --env-file deploy/compose/example/local/.env.example config --quiet` validates
- [x] `task dev:release:up` starts locally with TLS disabled — push, pull, and scan all verified
- [x] `task dev:release:up IMAGE_REPO=ttl.sh/8gcr/` pulls remote images
- [x] Token key generation produces PKCS#1 header (`RSA PRIVATE KEY`)
- [x] All images build and run as non-root
- [x] `task dev:up` dev environment starts, all services healthy — push, pull, and scan verified

PR #30, opened 2026-02-22
fix(devenv): use openssl genrsa -traditional to generate PKCS#1 RSA key

Aloui-Ikram:
## Problem

`task dev:gen:private-key` generates a PKCS#8 key (`BEGIN PRIVATE KEY`) 
but Harbor token service requires PKCS#1 (`BEGIN RSA PRIVATE KEY`), 
causing `docker login` to fail with 500 Internal Server Error.

## Change

In `taskfile/dev.yml`, changed:
```bash
openssl genpkey -algorithm RSA -out {{.TOKEN_KEY_FILE}} -pkeyopt rsa_keygen_bits:4096
```
to:
```bash
openssl genrsa -traditional -out {{.TOKEN_KEY_FILE}} 4096
```

## Testing

1. `task dev:clean`
2. `task dev:gen:private-key`
3. Verified `head -1 devenv/token_service_key.pem` shows `-----BEGIN RSA PRIVATE KEY-----`
```
root@PC:/home/ikram/harbor-next# head -1 devenv/token_service_key.pem
-----BEGIN RSA PRIVATE KEY-----
```
4. `docker login localhost:8080 -u admin -p Harbor12345` → `Login Succeeded`
```
root@PC:/home/ikram/harbor-next# docker login localhost:8080 -u admin -p Harbor12345
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
Login Succeeded
root@PC:/home/ikram/harbor-next#
```
FIxes #29

---
## Summary by cubic
Generate a PKCS#1 RSA private key in the dev environment to match Harbor token service requirements and stop docker login 500 errors. Replaced openssl genpkey with openssl genrsa -traditional 4096 in taskfile/dev.yml.

<sup>Written for commit 1112e3fdec17f36edb056c95c9ceb0321ce49864. Summary will update on new commits.</sup>

## Release Notes

* **Chores**
  * Updated development task configuration for improved key generation process.

PR #551, opened 2026-08-03
ci: bump delimit-action pin so fork-PR contributors can see schema-check output

infracore:
# ci: bump delimit-action pin so fork-PR contributors can see schema-check output

**Disclosure:** I am a maintainer of delimit-ai and the author of `delimit-ai/delimit-action`, the action this workflow already uses. This PR updates the pinned commit of **our own action** from a 2026-06-18 SHA to the current `main` HEAD. Nothing else in your workflow changes. If you'd rather not take the bump, that's completely fine — you can pin any SHA you prefer, or drop the step entirely; the schema-diff logic is also available in vendor-neutral form (any OpenAPI diff tool pointed at `api/v2.0/swagger.yaml` covers the basic breaking-change check).

## Why this matters to harbor-next specifically

Your current pin (`9cc52e7`, 2026-06-18) predates the fork-PR fix shipped in delimit-action#35 (2026-07-12). The action tries to post its report as a PR comment, but on PRs **from forks** the `GITHUB_TOKEN` is read-only for `pull-requests`, so the comment API returns 403 and — on your pinned version — the output is silently dropped. Since most harbor-next contributions come from forks, the check currently runs but shows contributors **nothing**: it is genuinely mute on the PRs where it matters most.

The new pin (`dc7cf46`) writes the report to the **GitHub job summary** as a fallback whenever the PR comment can't be posted, and correctly diagnoses the fork-403 case instead of swallowing it. Fork contributors will finally see the schema-diff output on the run's summary page.

## What changed between the SHAs (`9cc52e7..dc7cf46`, 18 commits)

Fixes relevant to you:
- **Job-summary fallback on fork PRs + correct 403 diagnosis** (#35) — the reason for this PR.
- **Template-safe `github-script` usage** (#37) — the comment step no longer interpolates report content into the script template; removes a script-injection class and adds a CI guard for it.
- **Context-aware breaking-change severity** (#30) — fewer false "breaking" classifications; severity now accounts for where in the spec a change occurs.
- **Crash fix in the diff evidence path** (#29) — non-string change details could crash report generation.
- **Fixed a dead Marketplace link in the runtime output** (#42).
- **README fixes + a README link-check CI** (#44, #45) — repo-internal, no runtime effect on consumers.

Changes you should know about before accepting (being upfront):
- The PR comment / job summary now **links to a hosted worked-example report on delimit.ai** and tags that link with a `?src=action-comment` attribution parameter (#32, #33, #43). This is a link in the report output, not telemetry from your CI — the action itself phones nothing home — but it does route readers to our site. If you'd prefer the old pin's plainer output, or want a flag to suppress the link, say so and I'll treat that as a feature request.
- The remainder is docs/README/marketplace copy and repo-internal CI (identity-guard workflows, tests) with no runtime effect on consumers.

Full comparison: https://github.com/delimit-ai/delimit-action/compare/9cc52e70501423dc6b55a8ffccec938920b04a2e...dc7cf46b8969c6e554e06ff1524f5e0271c62630

## Scope and posture

- One-line change: the pinned SHA in `.github/workflows/api-schema-check.yml` (the `# v1` comment convention is kept).
- No new permissions requested; the workflow's existing `contents: read` / `pull-requests: write` grant is unchanged.
- The action is advisory by design; if you set `continue-on-error: true` on the step it can never block a merge, and that remains true on the new pin. Happy to add that line to this PR if you want the advisory posture explicit.
- If maintainers prefer, I'm equally happy to close this and leave the pin as-is — it's your workflow.

### DIFF
--- a/.github/workflows/api-schema-check.yml
+++ b/.github/workflows/api-schema-check.yml
@@ -17,6 +17,6 @@ jobs:
         with:
           fetch-depth: 0
           persist-credentials: false
-      - uses: delimit-ai/delimit-action@9cc52e70501423dc6b55a8ffccec938920b04a2e # v1
+      - uses: delimit-ai/delimit-action@38251bbf68ff33094824e6fab9738556c3ccce0b # v1
         with:
           spec: api/v2.0/swagger.yaml


infracore:
Updated the pin to `38251bb` — the exact commit `v1` resolves to as of today — so the `# v1` comment is now literally accurate (addresses the zizmor/coderabbit/cubic findings: the earlier SHA was two commits behind the tag). Also amended with `Signed-off-by` per the contribution checklist. Between the previously-proposed SHA and this one, the only consumer-visible addition is a data-only report-artifact upload step (delimit-ai/delimit-action#46) that enables an optional fork-PR comment relay; nothing else in the runtime behavior changed. Full comparison: https://github.com/delimit-ai/delimit-action/compare/9cc52e70501423dc6b55a8ffccec938920b04a2e...38251bbf68ff33094824e6fab9738556c3ccce0b

bupd:
I am thinking of getting rid of this dependency - i dont think we need this here - since the upstream has this ??

infracore:
Totally your call — the PR description offers exactly that option and I mean it. One data point for the decision: the upstream copy only runs on goharbor/harbor's own PRs — it never sees harbor-next's. So the question is whether `api/v2.0/swagger.yaml` here ever changes outside upstream syncs. If it doesn't, dropping the step is reasonable and loses you nothing. If harbor-next carries its own API changes, the upstream run won't catch a breaking change introduced in this repo — that's the only thing the local copy adds. Happy to close this PR either way.

PR #659, opened 2026-08-20
fix: Carry X-Is-Resource-Name Header Into Audit Metadata

bupd:
Follow-up to #588, addressing cubic review findings surfaced on the 2.15 backport (#654) — the same code is on `main`, so it needs the fix here too.

- **P1**: `X-Is-Resource-Name` is a header parameter (see `api/v2.0/swagger.yaml`), not a query parameter, so the project audit resolver's `isResourceName(ce.RequestURL)` could never see it and still resolved all-digit project names as project IDs. The log middleware now captures the header into `commonevent.Metadata.IsResourceName` and the resolver uses that. (The removed query check read `x_is_resource_name`, a parameter the API never accepts.)
- **P3**: `pattern.ValidateKind` error message now lists empty as an accepted value (doublestar default), matching its actual behavior.

Same change applied to the upstream PR (goharbor/harbor#23762, second commit) and stacked on the 2.15 backport (#654).

### DIFF
--- a/src/controller/event/metadata/commonevent/model.go
+++ b/src/controller/event/metadata/commonevent/model.go
@@ -58,6 +58,9 @@ type Metadata struct {
 	ResponseCode int
 	// RequestURL request URL
 	RequestURL string
+	// IsResourceName indicates the request declared, via the X-Is-Resource-Name
+	// header, that name/ID path parameters are resource names
+	IsResourceName bool
 	// IPAddress IP address of the request
 	IPAddress string
 	// ResponseLocation response location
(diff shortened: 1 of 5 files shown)


bupd:
Backport to `release-2.15` is not needed — the `/backport v2.15` conflict is because this change already exists there.

#654 (`debbe350af0a`, the 2.15 backport of #588) was squash-merged **with the cubic review fixes already applied**, which is exactly what this PR ported forward to `main`. Verified: all 5 files touched by `b81d11c999` (`commonevent/model.go`, `pattern/matcher.go`, `auditext/event/project/project.go` + test, `middleware/log/log.go`) are byte-identical between `release-2.15` and this commit. Nothing to cherry-pick.

PR #687, opened 2026-08-24
fix(chart): extraTemplateManifests silently drops every document after the first

### LINKED ISSUE #667: bug(chart): extraTemplateManifests silently drops every document after the first
### Summary

`deploy/chart/templates/extraManifests.yaml` pipes each `extraTemplateManifests` entry through `fromYaml`, which parses only the **first** YAML document in a string. Every document after the first `---` separator is discarded silently — no error, no warning, and the rendered output looks well-formed.

Since `extraManifests` is the escape hatch for resources the chart does not model, a user grouping several related objects into one entry loses all but one of them, and will typically only find out when something downstream is missing.

### Cause

```
deploy/chart/templates/extraManifests.yaml:22
{{- $rendered := tpl . $ | fromYaml }}
```

`fromYaml` is single-document by definition. The multi-document case needs splitting before parsing.

### Reproduce

```yaml
extraTemplateManifests:
  - |
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: first
    ---
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: second
```

`helm template` emits only `first`. `second` is gone, with no diagnostic.

### Why it survived

`deploy/chart/tests/extramanifests_test.yaml` covers the single-document cases well — static pass-through, `tpl()` interpolation, and the empty case — but there is no multi-document case in the suite. A regression test feeding one entry containing two documents and asserting both are rendered would pin this.

### Two related problems in the same file

**A string entry in `extraManifests` fails hard.** The static branch does `deepCopy .` at line 10 and then reads `$manifest.metadata` at line 11, so a string entry aborts rendering with `wrong type for value; expected map[string]interface{}; got string`. Worth either accepting strings there for symmetry with `extraTemplateManifests`, or documenting that the static list is maps-only.

**Chomping after the separators.** Lines 10 and 22 open with `{{-` immediately after a `---` line, which trims the newline and yields `---apiVersion: v1` on a single line in the raw render. Currently harmless, because Helm's document splitter matches `(?m)^---`, but it is one edit away from breaking and makes the raw output awkward to read.

### Suggested fix

Split the rendered string on the document separator before parsing, and handle each document individually. `8gears/n8n-helm-chart` implements the same escape hatch and splits on `---` before templating, so it handles the multi-document case correctly — useful as a reference.

Worth noting the comparison runs the other way too: this chart's label handling is the better of the two. It uses `deepCopy` so it does not mutate `.Values`, and it merges user labels under `app.kubernetes.io/*`, where the n8n chart stamps the Helm 2 legacy scheme.

SurbhiAgarwal1:
Resolves issue #667.

### What this PR does / why we need it
This fixes a bug in deploy/chart/templates/extraManifests.yaml where multi-document YAML strings in extraTemplateManifests were not being properly split and handled correctly.

The changes also:
- Prevent newline chomping issues.
- Support string arrays safely in extraManifests.

Fixes #667.

### DIFF
--- a/deploy/chart/templates/extraManifests.yaml
+++ b/deploy/chart/templates/extraManifests.yaml
@@ -6,25 +6,48 @@ Extra manifests - escape hatch for any additional resources
 
 {{/* Static manifests - merged with chart labels */}}
 {{- range .Values.extraManifests }}
----
-{{- $manifest := deepCopy . }}
+{{- $docs := list . }}
+{{- if kindIs "string" . }}
+{{- $docs = regexSplit "(?m)^---" . -1 }}
+{{- end }}
+{{- range $doc := $docs }}
+{{- $manifest := $doc }}
+{{- if kindIs "string" $doc }}
+{{- if trim $doc }}
+{{- $manifest = fromYaml $doc }}
+{{- if $manifest.Error }}{{- fail (printf "invalid YAML in manifest: %s" $manifest.Error) }}{{- end }}
+{{- end }}
+{{- end }}
+{{- if and $manifest (kindIs "map" $manifest) $manifest.kind }}
+{{- $manifest = deepCopy $manifest }}
 {{- $userMeta := default dict $manifest.metadata }}
 {{- $userLabels := default dict $userMeta.labels }}
 {{- $mergedLabels := merge $userLabels $labels }}
 {{- $_ := set $userMeta "labels" $mergedLabels }}
 {{- $_ := set $manifest "metadata" $userMeta }}
-{{- toYaml $manifest }}
+---
+{{ toYaml $manifest }}{{ "\n" }}
+{{- end }}
+{{- end }}
 {{- end }}
 
 {{/* Templated manifests - can use .Values, .Release, etc. */}}
 {{- range .Values.extraTemplateManifests }}
----
-{{- $rendered := tpl . $ | fromYaml }}
-{{- $manifest := deepCopy $rendered }}
+{{- $rendered := tpl . $ }}
+{{- range $docStr := regexSplit "(?m)^---" $rendered -1 }}
+{{- if trim $docStr }}
+{{- $manifest := fromYaml $docStr }}
+{{- if $manifest.Error }}{{- $manifest = dict }}{{- end }}
+{{- if and $manifest (kindIs "map" $manifest) $manifest.kind }}
+{{- $manifest = deepCopy $manifest }}
 {{- $userMeta := default dict $manifest.metadata }}
 {{- $userLabels := default dict $userMeta.labels }}
 {{- $mergedLabels := merge $userLabels $labels }}
 {{- $_ := set $userMeta "labels" $mergedLabels }}
 {{- $_ := set $manifest "metadata" $userMeta }}
-{{- toYaml $manifest }}
+---
+{{ toYaml $manifest }}{{ "\n" }}
+{{- end }}
+{{- end }}
+{{- end }}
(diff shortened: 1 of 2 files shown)


Vad1mo:
thank you, we will fix that

bupd:
Ran the chart suite and then tried to break the new splitting logic. The core fix is right; one edge case regressed.

## What I ran

```
task helm:deps:build && task helm:unittest
Charts:      1 passed, 1 total
Test Suites: 23 passed, 23 total
Tests:       256 passed, 256 total
Snapshot:    106 passed, 106 total
```

(`main` baseline is 252; the 4 extra are yours.)

The diagnosis and the fix shape are both correct. On `main`, `tpl . $ | fromYaml` feeds a multi-document string into sprig's `fromYaml`, which wraps `yaml.Unmarshal` and cannot take more than one document in a stream. Templating the whole entry first, then splitting, then parsing each piece is the right order, and it preserves directives that span a `---` boundary. The tests genuinely cover that — the `RELEASE-NAME-first` / `RELEASE-NAME-second` cases are exactly what would regress if someone reintroduced per-document `tpl`.

## The regression

Both branches now gate on `{{- if and $manifest (kindIs "map" $manifest) $manifest.kind }}`. On `main` a map entry rendered unconditionally. Rendered the same values file against both:

```yaml
extraManifests:
  - apiVersion: v1
    metadata:
      name: no-kind-entry
    data:
      b: "2"
```

```
main          kind-less map entry -> RENDERED
this branch   kind-less map entry -> SILENTLY DROPPED (render succeeded, object absent)
```

So a user whose entry is missing `kind` — or misindented such that `kind` lands a level deeper — now gets nothing at all, with no error and a successful `helm template`. That is the same failure class the PR is fixing: a silently vanishing manifest. It should `fail` with a message naming the offending entry instead of skipping it.

## Error-handling asymmetry between the two branches

Static branch:
```
{{- if $manifest.Error }}{{- fail (printf "invalid YAML in manifest: %s" $manifest.Error) }}
```
Templated branch:
```
{{- if $manifest.Error }}{{- $manifest = dict }}{{- end }}
```

The templated branch swallows it. I confirmed an entry of `this is: [not valid yaml` renders successfully and produces nothing — on both `main` and this branch, so it isn't a regression, but the templated branch is the one users hit hardest (it's where `tpl` can produce malformed output) and it's the one that stays silent. Both should `fail`.

## One thing I checked and it's fine

I suspected `regexSplit "(?m)^---"` would split on a `---` inside a block scalar — e.g. a ConfigMap whose `data` value is itself a YAML document, a common escape hatch. It doesn't, because nested block-scalar content is always indented, so `^---` never matches at column 0. Rendered it both ways to be sure and the ConfigMap comes out byte-identical on `main` and here, `data` intact. Tightening to `(?m)^---\s*$` would still be cheap insurance for a top-level entry, but it isn't a live bug.

## Test gaps

- no `hasDocuments: count:` assertion on the **static** multi-doc path (only per-index names)
- no case for a map entry without `kind`
- no case for invalid YAML in `extraTemplateManifests`

## Not tested

Nothing applied to a live cluster, and no upgrade path from a release created before this change. Also note the branch is 73 commits behind `main`, so I tested the head as-is rather than the merge result.

Schema is fine, for the record: `values.schema.json` declares both keys as bare `{"type": "array"}` with no `items` constraint, so the new string-entry support needs no schema change and `task helm:schema-check` isn't affected. The `description` ("Extra static manifests to deploy") no longer quite matches the widened contract though.

PR #691, opened 2026-08-24
ci: Accept Hierarchical Commercial Branch Names In Release Notes

bupd:
## Problem

`.github/scripts/update-release-notes.sh` rejects any commercial patch
branch name containing `/`:

```sh
if [[ "${branch}" == */* || "${branch}" == *..* ]]; then
  echo "Invalid commercial branch name in series: ${branch}" >&2
  exit 1
fi
```

This predates the `release-2.15/000N-*` naming convention that
`taskfile/commercial-patches` on `release-2.15` actually uses. Every
release-notes run for v2.15.7 has failed:

```
Invalid commercial branch name in series: release-2.15/0001-commercial-feature-gate
```

See: https://github.com/container-registry/harbor-next/actions/runs/32698718386/job/97351848443

## Fix

Validate with `git check-ref-format --branch` instead, matching how
`taskfile/release-ready.yml`'s own `_validate-series` task already
validates the same manifest. This correctly accepts hierarchical branch
names while still rejecting genuinely malformed ones (path traversal,
`..` ranges, etc.) — the actual injection concern the original check
was going for.

## Verification

```sh
git check-ref-format --branch "release-2.15/0001-commercial-feature-gate"  # accepted
git check-ref-format --branch "../evil"                                    # rejected
git check-ref-format --branch "foo..bar"                                   # rejected
```

### DIFF
--- a/.github/scripts/update-release-notes.sh
+++ b/.github/scripts/update-release-notes.sh
@@ -73,6 +73,10 @@ node .github/scripts/format-release-notes.mjs \
 
 if [[ -n "${preview_pr_number}" ]]; then
   release_branch="${GITHUB_REF_NAME:?GITHUB_REF_NAME is required for a release PR preview}"
+elif [[ "${GITHUB_REF_TYPE:-}" == "branch" && -n "${GITHUB_REF_NAME:-}" ]]; then
+  # push-triggered path only. Tag-triggered runs (manual recreate on a
+  # tag ref) have GITHUB_REF_NAME=<tag>, not a branch, so fall through.
+  release_branch="${GITHUB_REF_NAME}"
 else
   release_branch=$(github_retry gh release view "${TAG_NAME}" \
     --repo "${GITHUB_REPOSITORY}" \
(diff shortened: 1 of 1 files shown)


bupd:
While verifying this fix locally I found a second, related bug in the same script: the 'verify an image signature' cosign snippet in generated release notes referenced the wrong workflow file (`publish-images.yml`, never actually invoked for real releases) and derived the ref from `gh release view`'s `targetCommitish`, which returns a raw commit SHA instead of a branch name for release-please-created releases. Both combine to produce a `--certificate-identity` that never matches the actual signing cert — the copy-pasted verify command always fails.

Images are actually signed by `release-please.yml`'s own `sign` job, confirmed directly:

```
cosign verify --certificate-identity="https://github.com/container-registry/harbor-next/.github/workflows/release-please.yml@refs/heads/release-2.15" ... 8gears.container-registry.com/8gcr/harbor-core:v2.15.7
# → succeeds
cosign verify --certificate-identity="https://github.com/container-registry/harbor-next/.github/workflows/publish-images.yml@refs/heads/<sha>" ...
# → fails: SAN mismatch
```

Fixed both in the second commit: point at `release-please.yml`, and prefer `GITHUB_REF_NAME` (correctly set by the calling workflow) over `targetCommitish` for the branch name.

bupd:
good catch. fixed: gate GITHUB_REF_NAME on GITHUB_REF_TYPE==branch, else fall to targetCommitish. tag-triggered manual recreate now falls through correctly.

PR #699, opened 2026-08-24
chore: Fall Back To Plain Binary When Delve Is Missing In Dev Env

bupd:
Follow-up to #698 (this part missed the merge). When `/go/bin/dlv` is absent from the dev image, every air hot-reload restart dies with `/bin/sh: /go/bin/dlv: not found` — the service stays down and the portal dev proxy answers 504 for every request until the image is rebuilt.

The three air configs now check for dlv in `full_bin` and fall back to executing the freshly built binary directly. A broken or missing debugger install degrades step-through debugging but can never take Core, JobService, or RegistryCtl down.

Verified locally: with dlv removed from the container, hot reload restarts keep the services serving.

### DIFF
--- a/devenv/air.core.toml
+++ b/devenv/air.core.toml
@@ -15,7 +15,8 @@ tmp_dir = "tmp"
 
   # Delve debugger on port 2345, starts immediately (--continue)
   # Attach anytime with: dlv connect localhost:2345
-  full_bin = "/go/bin/dlv exec ./tmp/core --listen=0.0.0.0:2345 --headless=true --api-version=2 --accept-multiclient --continue --"
+  # Falls back to the plain binary when dlv can't run; run() keeps Air-appended args working.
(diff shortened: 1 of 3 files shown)


bupd:
Addressed all bot findings in `9ea420e`:

- **cubic P1 (valid, the important one):** registryctl's `args_bin = ["-c", "config/registryctl.yml"]` gets appended by Air after `full_bin`, which with the previous `if …; fi` produced `fi -c config/registryctl.yml` — a shell syntax error that would have kept RegistryCtl down even with a healthy dlv. All three configs now wrap the logic in a `run()` function so appended args land as function arguments in both branches.
- **cubic P2 / coderabbit:** the guard is now `dlv version >/dev/null 2>&1` instead of `[ -x … ]`, so an existing-but-broken dlv also triggers the fallback.
- **cubic P3:** the fallback logs `air: dlv unavailable - running plain binary (no debugger on :PORT)` to stderr.

Verified live in the dev environment: with dlv absent, Core and RegistryCtl both come up on the plain binary (RegistryCtl booting confirms the appended `-c config/registryctl.yml` survives), and the notice appears in the container logs.

PR #709, opened 2026-08-25
fix: Validate and truncate client IP/user agent before audit capture

bupd:
## Summary

Fixes two review findings left unaddressed on #701 before it merged (cubic-dev-ai + coderabbitai both flagged the same underlying gap):

1. **Silent audit-record loss.** `X-Forwarded-For`/`User-Agent` were captured and persisted verbatim into `audit_log_ext`, with no length bound. A header longer than the column widths (`client_address varchar(255)`, `user_agent varchar(1024)`) made the audit insert fail; the triggering action still succeeded, so it shipped with **no audit record at all** and only a swallowed error in the logs.
2. **Unvalidated / spoofable client IP.** `X-Forwarded-For` was trusted as-is and stored verbatim — a client could send an arbitrary non-IP string, or a spoofed address, straight into the security audit log.

`GetClientIP` (`server/middleware/security/basic_auth.go`) now takes only the first hop of a comma-separated `X-Forwarded-For` and requires it to actually parse as an IP, falling back to `RemoteAddr` otherwise. Both `GetClientIP` and `GetUserAgent` cap their result to the `audit_log_ext` column widths, truncating by **rune count**, not byte count, so a multi-byte character can't be split mid-sequence — matching how Postgres itself counts `varchar(n)`.

This is the single shared capture point used by every consumer added in #701/#704 (the generic REST audit pipeline, and artifact push/pull/delete via the registry and proxy-cache paths), so the fix covers all of them at once — no other files needed to change.

Considered and deliberately out of scope: fully preventing IP spoofing would require a trusted-proxy allowlist for `X-Forwarded-For`, which doesn't exist anywhere in Harbor today and is a materially larger change. This PR closes the concrete "garbage/oversized header breaks or pollutes the audit log" gap the reviews flagged; genuine spoofing-when-there's-no-trusted-reverse-proxy-in-front is a pre-existing deployment-level gap, unchanged by this PR.

## Related Issues

Follow-up to review feedback on #701 (merged before it could be addressed).

## Type of Change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Documentation (`docs:`)
- [ ] Tests (`test:`)

## Release Notes

Fixes a bug where an unusually long `X-Forwarded-For`/`User-Agent` header could silently prevent an audit log entry from being written, and hardens client-IP capture against non-IP header values.

## Testing

Added cases to the existing `TestGetClientIP`/`TestGetUserAgent` table tests (proxy chain first-hop, whitespace, IPv6, garbage-header fallback, oversized-input truncation) plus a new `TestTruncateRunes` covering multi-byte input. `gofmt -l` and `go vet -tags db ./server/middleware/security/...` are both clean. I was not able to execute this package's `db`-tagged tests in this sandbox — its `TestMain` requires a live Postgres already migrated to a matching schema version, and the only instance available locally was on a newer version — so correctness was verified by build/vet plus hand-tracing against the test table.

## Checklist


src/server/middleware/security/basic_auth.go
```diff
@@ -29,6 +31,12 @@ import (
 
 type basicAuth struct{}
 
+// mirror audit_log_ext.client_address/user_agent column widths
```
bupd:
```suggestion
// mirror audit_log_ext.client_address/user_agent column widths in db
```

bupd:
Both fixed:

1. **Port-qualified first hop (P2)** — this was a real regression. `GetClientIP` now strips a `host:port`/`[host]:port` wrapper (`net.SplitHostPort`) before `net.ParseIP`, so `X-Forwarded-For: 1.1.1.1:8080, ...` resolves to `1.1.1.1` instead of silently falling back to `RemoteAddr`. Added IPv4 and IPv6 (bracketed) port-qualified cases to the table tests.
2. **Tests stuck behind `db` tag (P3)** — moved `TestGetClientIP`/`TestGetUserAgent`/`TestTruncateRunes` (plus a new `TestParseIP`) into a new `client_info_test.go` without the build tag, keeping only the two DB-backed auth tests in `basic_auth_test.go`. They now actually execute on a plain `go test ./server/middleware/security/...` — verified locally, 13/13 passing.

(Worth noting: this package is excluded from both the pure and db lanes in `taskfile/test.yml`'s CI config already, for unrelated cert-file reasons predating this PR — so this specifically fixes local/IDE test runs and any future change to that exclusion list, not today's CI. Still the right fix.)

bupd:
Fixed the Go Lint CI failure: revive flagged `func truncateRunes(s string, max int) string` for shadowing the `max` builtin (`redefines-builtin-id`). Renamed the parameter to `maxLen`. Verified clean with `golangci-lint run ./server/middleware/security/...` (v2.12.2, matching CI) — 0 issues.

Issue #710, opened 2026-08-25
Authoritative DB migration: Way 1 vs Way 2 benchmark results


bupd:
# Authoritative DB migration: Way 1 vs Way 2 benchmark results

## Background

Harbor Next's numbered migrations (`make/migrations/postgresql/NNNN_*.up.sql`) share one number
space with upstream goharbor/harbor, tracked by `golang-migrate` as a single "current version"
integer — not content, not checksums. Harbor Next previously shipped its own `0181`/`0182` on a
`release-2.15` branch (branding, identity providers). Upstream has since shipped its **own**
`0181`/`0182` in later 2.15.x patches. Because the migrator only checks "have I passed this
number," an install that already applied Harbor Next's old `0181`/`0182` will **skip** upstream's
real `0181`/`0182` on upgrade — silently losing schema changes. This is a structural collision
between two independently-numbered migration streams sharing one space, and it will keep
recurring.

**Fix direction**: stop injecting Harbor Next-owned numbers into that shared space, permanently.
Move all Harbor Next-specific schema needs into a separately-tracked, idempotent mechanism
decoupled from `golang-migrate`'s counter. Two designs were built and benchmarked:

- **Way 1** (`experiment/authoritative-db-single-file`) — one evergreen `harbor_next.sql`,
  reconciled idempotently after every startup, not tracked in `schema_migrations` at all. Verified
  to already be the exact idempotent union of the old colliding `0181`/`0182` DDL, so it fully
  self-heals any install regardless of migration-number history.
- **Way 2** (`experiment/authoritative-db-versioned-file`) — `harbor_next_<version>.sql` per
  release, tracked in a small `harbor_next_schema_migrations` table, only applying files newer
  than the last recorded version (skips entirely once caught up).

## Methodology

**Environment note**: the target real k3d + Helm chart 3-replica deployment wasn't achievable —
rootless podman in this environment doesn't delegate the `cpuset` cgroup controller that k3s's
server requires unconditionally (`failed to find cpuset cgroup (v2)`), and fixing it needs a
`sudo` systemd drop-in plus a session restart, which was out of scope to do unattended. Instead,
this benchmark uses a 3-replica `podman-compose` stack (`postgres` + `redis` + 3× `core`, no
portal/registry/jobservice — core doesn't dial those at startup) — the concurrent-migration /
advisory-lock behavior under test is identical to a real k8s deployment; what's missing is
[...]

PR #711, opened 2026-08-25
feat: Add authoritative Harbor Next schema reconciliation

bupd:
## Problem

Harbor Next's numbered migrations (`make/migrations/postgresql/NNNN_*.up.sql`) share one number
space with upstream goharbor/harbor. `golang-migrate` tracks only a single "current version"
integer, not content. Harbor Next previously shipped its own `0181`/`0182` on `release-2.15`
(branding, identity providers); upstream has since shipped its **own** `0181`/`0182` in later
2.15.x patches. An install that already applied Harbor Next's old `0181`/`0182` will **skip**
upstream's real ones on upgrade, silently losing schema changes. This will keep recurring as long
as two independently-numbered migration streams share one space.

## Fix

Stop injecting Harbor Next-owned numbers into that shared space, permanently. `harbor_next.sql` is
a separately-tracked, idempotent schema file reconciled after every startup — not tracked in
`schema_migrations` at all, guarded by a Postgres advisory lock so concurrent core replicas don't
race. It's the exact idempotent union of the old colliding `0181`/`0182` DDL (branding,
identity_providers, robot_identity_providers, claim_rules), so it fully self-heals any install
regardless of migration-number history — verified against the actual historical files from
`release-2.15/0002-branding` and `release-2.15/0004-identity-providers`.

## Benchmark

Compared against a versioned-file alternative and a no-mechanism baseline under a real 3-replica
concurrent-migration load test. Request latency is indistinguishable across all three; this
mechanism's steady-state overhead on pod restart is ~3.4ms. Full results, methodology, and known
limitations: https://github.com/container-registry/harbor-next/issues/710

## Release Notes

Harbor Next's own schema additions (branding, workload identity federation) are now reconciled
via a single idempotent file after every migration, decoupled from upstream's numbered migration
sequence — fixing a class of bug where a future upstream migration could silently be skipped
because its number collided with one Harbor Next had already used internally.

### DIFF
--- a/make/migrations/postgresql/harbor_next.sql
+++ b/make/migrations/postgresql/harbor_next.sql
@@ -43,7 +43,7 @@ CREATE TABLE IF NOT EXISTS identity_providers (
 CREATE TABLE IF NOT EXISTS robot_identity_providers (
     id                   SERIAL PRIMARY KEY,
     identity_provider_id INT NOT NULL REFERENCES identity_providers(id) ON DELETE CASCADE,
-    robot_id             INT NOT NULL REFERENCES robot(id) ON DELETE CASCADE,
+    robot_id             BIGINT NOT NULL REFERENCES robot(id) ON DELETE CASCADE,
     creation_time        TIMESTAMP DEFAULT NOW(),
     UNIQUE (identity_provider_id, robot_id)
 );
@@ -51,7 +51,7 @@ CREATE TABLE IF NOT EXISTS robot_identity_providers (
 CREATE TABLE IF NOT EXISTS claim_rules (
     id                   SERIAL PRIMARY KEY,
     identity_provider_id INT NOT NULL REFERENCES identity_providers(id) ON DELETE CASCADE,
-    robot_id             INT NOT NULL DEFAULT 0,
+    robot_id             BIGINT NOT NULL DEFAULT 0,
     claim_path           TEXT NOT NULL,
     value                TEXT,
     creation_time        TIMESTAMP DEFAULT NOW()
--- a/make/migrations/postgresql/harbor_next.sql
+++ b/make/migrations/postgresql/harbor_next.sql
@@ -7,6 +7,10 @@
 --
 -- Only additive, data-preserving changes belong here. Destructive changes and
 -- large data backfills require a separately reviewed operational procedure.
+--
+-- branding and identity_providers/robot_identity_providers/claim_rules were
+-- formerly release-2.15 migrations 0181/0182; both numbers were later reused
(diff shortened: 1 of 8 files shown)


bupd:
While reviewing the patches that need this schema (\`0002-branding\`, \`0004-identity-providers\`), found they'd both drifted into shipping their own numbered migrations again — \`0192_8gears_2.15_branding.up.sql\` and \`0193_8gears_2.15_identity_providers.up.sql\` — the exact same collision-prone pattern this PR exists to eliminate, just with new numbers. DDL was statement-for-statement identical to what's already in \`harbor_next.sql\` here.

Removed both numbered files (pushed directly to \`8gcr/0002-branding\` and \`8gcr/0004-identity-providers\`, fast-forward only, no history rewrite) — they're fully superseded by this PR's authoritative reconciliation.

bupd:
Went through both automated reviews (CodeRabbit, cubic). Fixed the two real findings:

- **`robot_id` INT→BIGINT** (cubic #2/#3): confirmed — `0190_2.16.0_schema.up.sql` widens `robot.id` to bigint. `robot_identity_providers.robot_id` and `claim_rules.robot_id` recreated the overflow risk. Fixed, plus updated the DB test's `robot` stub to `BIGSERIAL` to match, and reran the integration test against real Postgres.
- **`NewMigrator` ignoring `database.URL`** (cubic #4): confirmed — pre-existing bug in `dao.NewMigrator`, made consequential by this PR since there are now two migration phases that need to agree on target DB. With `POSTGRESQL_URL` set (e.g. RDS IAM auth via `0006-aws-rds-iam-auth`), the numbered-migration phase would silently target a different database than the authoritative-schema phase. Fixed with a unit-tested `migratorDSN()` helper.

Not acting on the rest, with reasons:

- **"IF NOT EXISTS doesn't reconcile existing objects"** (CodeRabbit + cubic #1): by design — the file's own header already states "Only additive, data-preserving changes belong here. Destructive changes and large data backfills require a separately reviewed operational procedure." These tables are net-new (no prior release shipped a different shape), so there's nothing to reconcile today; evolving an already-shipped table's shape is explicitly out of scope for this mechanism.
- **VERSION bump**: this repo's `main` uses `always-bump-minor` via release-please automation — `VERSION` isn't meant to be hand-edited in a PR.
- **Test synchronization barrier** (cubic #5, P3): the existing test already proves the practical property that matters — concurrent + repeated application converges to a consistent schema. Not adding a fault-injected race-window assertion for now.
- Docstring coverage: this codebase doesn't have an 80% docstring convention; not retrofitting one here.

PR #722, opened 2026-08-25
fix: Use canonical Apache license URL in authoritative.go header

Vad1mo:
goheader expects `http://www.apache.org/licenses/LICENSE-2.0` in the file header. #711 introduced `src/migration/authoritative.go` with the `https://` form, so `Go Lint` now fails on every PR targeting `main` (e.g. #716, #720). One-character fix.


bupd:
This fixes `authoritative.go` itself, but the two test files added alongside it in #711 have the identical goheader mismatch and will keep failing Go Lint after this merges — opened #726 for those.

PR #724, opened 2026-08-25
ci: Run workflows on GitHub-hosted ubuntu-26.04 runners

bupd:
## Summary

Switches all workflows from the `vars.RUNNER`-driven runner selection (self-hosted, falling back to `ubuntu-latest`) to a hardcoded GitHub-hosted `ubuntu-26.04` runner.

## Related Issues

N/A

## Type of Change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Documentation (`docs:`)
- [ ] Tests (`test:`)

## Release Notes

N/A (CI-only change, excluded from release notes)

## Testing

- [ ] Unit tests added/updated

All 25 modified workflow files verified as valid YAML (`yaml.safe_load`). No `vars.RUNNER` references remain. CI running this PR itself is the real-world test.

## Checklist

### DIFF
--- a/.github/workflows/backport-suggestion.yml
+++ b/.github/workflows/backport-suggestion.yml
@@ -16,7 +16,7 @@ jobs:
       ${{
         github.event.pull_request.base.ref == 'main'
       }}
-    runs-on: ${{ vars.RUNNER || 'ubuntu-latest' }}
+    runs-on: ubuntu-26.04
     steps:
       - name: Suggest backports
         uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -17,7 +17,7 @@ jobs:
         startsWith(github.event.comment.body, '/backport ') &&
         contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)
       }}
-    runs-on: ${{ vars.RUNNER || 'ubuntu-latest' }}
+    runs-on: ubuntu-26.04
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       GH_REPO: ${{ github.repository }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ concurrency:
 jobs:
   build:
     name: Build Binaries
-    runs-on: ${{ (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository || contains(fromJSON('["OWNER","MEMBER"]'), github.event.pull_request.author_association)) && vars.RUNNER || 'ubuntu-latest' }}
+    runs-on: ubuntu-26.04
     steps:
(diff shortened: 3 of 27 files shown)


bupd:
Addressed the two real findings, and explaining the other two:

**Fixed:**
- **`RELEASE.md` still documents `RUNNER`** — dropped that row from the configuration table now that nothing reads it.
- **`cache-cleanup.yml`'s `cleanup-host` job is now a no-op** — correct catch. It trimmed a persistent self-hosted `$GOCACHE`/`$GOMODCACHE` under `/home` or `/opt/ci-cache`, which doesn't exist once every job runs on an ephemeral GitHub-hosted runner. Removed the job, kept `cleanup-github-cache` (operates on the Actions cache API, runner-agnostic).

**Intentional, not fixing:**
- **Dropping the `vars.RUNNER` override / pinning to `ubuntu-26.04`** — this is the actual goal of the PR: move CI off self-hosted infrastructure onto GitHub-hosted runners. Reintroducing `vars.RUNNER || 'ubuntu-26.04'` would silently preserve the self-hosted path this PR is meant to retire. Point about GitHub eventually deprecating versioned runner images (as with `ubuntu-20.04`/`22.04`) is fair — that'll be a future one-line bump across these files when it happens, not a reason to keep the override today.

Also fixed the pre-existing `goheader` failure in `migration/authoritative.go` that was failing Go Lint on this and presumably every other open PR (license URL used `https://` with no indent instead of the repo's `http://` + 4-space-indent convention) — unrelated to this PR's scope but blocking its CI, so bundled it in.

Verified with the exact CI linter version (`golangci-lint v2.12.2 run ./...`) — 0 issues — and re-validated all 25 modified workflow YAML files parse cleanly.

bupd:
Fixed a real regression I introduced: restored the trust-gated \`vars.RUNNER\` conditional for the 8 workflows that had it (\`build\`, \`test\` (x2), \`lint-go\`, \`lint-ui\`, \`lint-actions\`, \`spellcheck\`, \`dependency-review\`, \`vulnerability-check\`) — several of these touch the commercial patch build path and are triggered by \`pull_request\`, so the OWNER/MEMBER check exists to keep self-hosted runner access away from arbitrary external-fork PRs.

My first pass collapsed that gate along with the plain \`vars.RUNNER || 'ubuntu-latest'\` fallback pattern used everywhere else, which would've meant *nobody* — including trusted internal PRs — got \`vars.RUNNER\` anymore on those 8 workflows.

Now: those 8 keep \`... && vars.RUNNER || 'ubuntu-26.04'\` (trusted actors still get the self-hosted runner when configured; forks still get a GitHub-hosted sandbox, just ubuntu-26.04 instead of ubuntu-latest). The other 17 workflows (release-please, publish-*, backport, cache-cleanup, etc. — none reachable by an arbitrary external PR) move to \`ubuntu-26.04\` unconditionally, as before.

bupd:
Confirmed — removed `vars.RUNNER ||` entirely, including from the 8 previously trust-gated workflows. Every workflow now runs on GitHub-hosted `ubuntu-26.04` unconditionally, no self-hosted runner anywhere and no OWNER/MEMBER branching left to maintain. Pushed.

PR #765, opened 2026-08-29
ci: Run Renovate daily and let it update its own version

Vad1mo:
## Problem

Two defects in how the Renovate runner is scheduled and versioned, both found by running renovate 44.46.5 locally against this repo.

The workflow cron was `0 1 */2 * *`. A day-of-month step lands on odd dates, and four packageRules in `renovate.json` are scheduled `* 0-3 * * 1`, Mondays only: Go majors, versions.env majors, GitHub Actions and Docker. A Monday has to coincide with an odd date for any of them to be evaluated. Over the coming year that is 26 of 52 Mondays, with gaps of up to 21 days, so those four rules ran at roughly half their intended cadence and irregularly.

`renovate-version: 44.46.5` was a bare workflow input carrying no `# renovate:` annotation, and there were no such annotations anywhere in `.github/workflows/`. The `renovatebot/github-action` SHA is kept current by the github-actions manager, but the Renovate version it installs was frozen at whatever was typed in #750. It is now five minors behind.

## Change

`cron: "0 1 * * *"`. The schedules in `renovate.json` all sit in the 0-3 UTC window, so the workflow only has to open that window daily and let the config own the cadence. That is what those schedules are for.

`RENOVATE_VERSION` moves to `versions.env` with a `datasource=npm depName=renovate` annotation, picked up by the custom manager already defined in `renovate.json` and restored in #763. The workflow gains a checkout plus the same `Load versions` step `pr-ci.yml` uses, and reads `${{ env.RENOVATE_VERSION }}`.

## Consequences

The runner goes from every other day to daily. On a public repo the Actions minutes are free, and a run is roughly 3 minutes.

Renovate now proposes its own upgrades, so a bad Renovate release can arrive by PR. The 7 day cooldown from #763 applies to it like any other dependency, and the PR is reviewable before merge.

## Verification

Local dry run (`--platform=local`) against this branch, renovate 44.46.5:

- Extraction picks up `depName: renovate` from `versions.env` via the regex manager.
- Lookup proposes `44.46.5 -> 44.51.0`, `updateType: minor`.
- zizmor is unchanged from `main` on this file: no findings, 2 suppressed.

### DIFF
--- a/.github/workflows/renovate.yml
+++ b/.github/workflows/renovate.yml
@@ -2,9 +2,11 @@ name: Renovate
 
 on:
   schedule:
-    # 01:00 UTC on odd days of month (~every 2 days), inside the 0-3 UTC
-    # windows in renovate.json; consecutive runs at 31st -> 1st are fine
-    - cron: "0 1 */2 * *"
+    # Daily at 01:00 UTC. Cadence is owned by the schedules in renovate.json,
+    # which all sit in the 0-3 UTC window; this only has to open that window.
+    # A day-of-month cron (*/2) lands on odd dates, so the Monday-scheduled
+    # rules there would have fired on only half the Mondays in a year.
+    - cron: "0 1 * * *"
   workflow_dispatch:
 
 permissions: {}
@@ -18,6 +20,13 @@ jobs:
     runs-on: ubuntu-26.04
     timeout-minutes: 60
     steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Load versions
+        run: grep -E '^[A-Z_]+=' versions.env >> "$GITHUB_ENV"
+
       # Dedicated Renovate app, not 8gcr-sync: the sync app declares only
       # actions/checks/statuses read + contents/pull-requests write, so the
       # downscope below can never be satisfied by it. The app installed for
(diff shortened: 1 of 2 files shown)


Vad1mo:
Both review points addressed in b51f05bb7.

**contents: read (cubic P1, Copilot)** — valid and would have broken every scheduled run. The workflow sets `permissions: {}` and the job declared none, so `GITHUB_TOKEN` had no scopes and the new checkout would have failed before Renovate started. Every other checkout-using workflow in this repo declares `contents`. Added at job level with a comment noting Renovate itself acts through the app token, never `GITHUB_TOKEN`.

**Scoping the versions.env read (cubic P3)** — valid, and taken further than suggested. Rather than narrowing the grep to `RENOVATE_VERSION`, the value is now read as a step output so nothing enters the job environment at all:

```yaml
- name: Read Renovate version
  id: versions
  run: |
    version="$(grep -E '^RENOVATE_VERSION=' versions.env | cut -d= -f2)"
    test -n "${version}" || { echo "::error::RENOVATE_VERSION missing from versions.env"; exit 1; }
    echo "renovate=${version}" >> "$GITHUB_OUTPUT"
```

The suggested grep would still have exported `RENOVATE_VERSION` itself into the Renovate process. Renovate reads `RENOVATE_*` as config overrides (that is how `RENOVATE_PLATFORM` and `RENOVATE_DRY_RUN` work in this same file), so a step output closes the class rather than shrinking it. A local run with `RENOVATE_VERSION` in the environment produced no warning today, so the current exposure was benign, but the mechanism is real. The empty check also turns a typo in versions.env into a clear failure instead of an empty `renovate-version` input.

Verified end to end, not just by inspection: dispatched this branch with a temporary `RENOVATE_DRY_RUN: full` (run 33254812792). Checkout, `Read Renovate version`, app token mint and Renovate all succeeded, the log confirms `renovateVersion: 44.46.5` resolved from the step output, and the dry run reported `DRY-RUN: Would commit files ...` without writing anything. The temporary flag and its revert were dropped from history afterwards; the branch is back to the reviewed tree.

PR #773, opened 2026-08-30
fix: Reject requests with 429 on DB connection pool exhaustion

bupd:
image
 
Closes the production failure mode where an exhausted DB connection pool turns into a system-wide hang: requests queue forever inside `pgxpool.Acquire` (Beego ORM issues queries with `context.Background()`, so no client-side deadline ever fires), nothing logs, and the system cannot recover on its own.

## What changed

**1. Reject instead of queueing forever (core HTTP layer — allowlisted user-facing routes only)**
New `server/middleware/dbadmission` middleware, installed ahead of the orm/transaction middlewares. It caps in-flight requests at **pool size × 3** — up to pool size holding connections plus **2x pool size** waiting for one. The slot is reserved atomically *before* any check, so a concurrent burst cannot observe spare capacity and slip past the bound (race flagged by review, fixed). Beyond the cap, requests are rejected immediately with **429 Too Many Requests** (with `Retry-After: 1`) and a `DB connections exhausted` warning log (method, path, conns in use, in-flight count).

**Deliberately scoped as an opt-in allowlist** (`dbAdmissionGated` in `core/middlewares`): Harbor's DB pool is shared by many internal callers — jobservice status callbacks under `/service/notifications`, GC and scan hooks, registry token auth, the `/v2/` data plane — and not all of them handle an unexpected 429 safely (a mishandled response in a GC flow can be destructive). The gate therefore starts with interactive user-facing requests only:

- `POST /c/login`
- `POST /api/v2.0/projects` (create project)
- `POST /api/v2.0/users` (create user)
- `POST /api/v2.0/robots` (create robot account)

Every other route bypasses the gate. The list grows one route at a time, each addition only after verifying all callers of that route back off correctly on 429.

**2. Bounded waiter queue**
pgxpool exposes no waiter count, so the middleware tracks admissions with its own atomic counter for the request lifetime. Waiter allowance = `2 * MaxConns` (e.g. 10 connections → max 20 queued requests).

**3. Kill stuck/slow queries**
Server-side `statement_timeout` is now set on every pooled connection via pgx `RuntimeParams` — the only reliable lever, since Beego drops the request context. Configurable via `POSTGRESQL_STATEMENT_TIMEOUT` (default **5m**, negative disables). Applied in `lib/dbpool.applyPoolConfig`, so core, jobservice and exporter all get it, and wired through the Helm chart as `database.statementTimeout` (core configmap + exporter env). Guardrails: values above PostgreSQL's int32-ms ceiling (~24.8 days) fail fast at startup; positive sub-millisecond values round up to 1ms instead of silently disabling; a `statement_timeout` already present in a URL-style DSN takes precedence. The authoritative schema migration runs with `SET LOCAL statement_timeout = 0` so long DDL (index builds on large tables) is not killed mid-upgrade; numbered migrations use their own connection and are unaffected.

**4. Health check no longer hangs**
`databaseHealthChecker` now uses the previously unused `Pool.Healthy()` (pool ping, 5s timeout) instead of `SELECT 1` through the ORM, which blocked indefinitely on an exhausted pool.

## Tests
- `lib/dbpool`: statement_timeout default / explicit / negative-disables / URL-param-precedence / sub-ms round-up / above-max rejection; live-Postgres integration tests (`-tags db`) proving a stuck `pg_sleep(10)` is killed in ~0.5s with SQLSTATE 57014 and the connection is reusable.
- `server/middleware/dbadmission` (incl. `-race`): uninitialized pool pass-through, sequential admission, in-flight cap → 429 + `Retry-After` + slot release, **burst-race regression** (stats reporting spare capacity must not bypass the bound), 200-goroutine burst hammer (peak in-flight never exceeds cap, no leaks), skipper bypass while gate is full.
- `core/middlewares`: allowlist skipper table test — gated routes vs internal paths (`/service/notifications/...`, `/service/token`, `/v2/` manifests/blobs, ping, health) that must never be gated.
- `migration`: SET LOCAL ordering + error propagation.
- Helm: unittest 249/249, snapshots updated, README regenerated with pinned helm-docs, values.schema.json extended.

## Release Notes
When the database connection pool is exhausted, Harbor now rejects excess **user-facing API requests** (login, project/user/robot creation) with HTTP 429 and a `Retry-After` header (bounded waiter queue of 2x the pool size) instead of queueing them indefinitely, and logs "DB connections exhausted". Internal and registry data-plane traffic is never rejected by this gate. A new server-side statement timeout (`POSTGRESQL_STATEMENT_TIMEOUT` / Helm `database.statementTimeout`, default 5m) kills stuck or slow queries so their connections return to the pool. **Upgrade note:** the statement timeout is active by default after upgrading — statements running longer than 5 minutes are cancelled by PostgreSQL (schema migrations are exempt). If you rely on longer-running statements (e.g. very large retention or audit-log purges), raise the timeout or disable it with `POSTGRESQL_STATEMENT_TIMEOUT=-1s`. The database health check now uses a bounded pool ping instead of hanging when the pool is saturated.


bupd:
Addressed the review findings in 49031eb:

- **Burst race past the waiter bound (cubic P1, coderabbit)** — valid. Redesigned: the in-flight slot is now reserved atomically before any stats check, capping in-flight requests at pool size × 3 unconditionally. Added a `-race` regression test where stats report spare capacity but the bound still holds.
- **statement_timeout > ~24.8 days breaks new connections (cubic P1)** — valid. Values above PostgreSQL's int32-ms ceiling now fail fast with a config error at pool creation.
- **Sub-millisecond timeout truncates to 0 = disabled (cubic P2, coderabbit)** — valid. Positive values now round up to 1ms; test added.
- **Default 5m silently active on upgrade, may kill legitimate long statements (cubic P2)** — partially accepted. Killing long-stuck statements is the point of this change, so the default stays on, but: the authoritative schema migration now runs with `SET LOCAL statement_timeout = 0` (long DDL is safe; numbered migrations use a separate connection and were never affected), and the upgrade behavior change is called out in the Release Notes with the `-1s` opt-out.
- **Log printed the cap, not the live waiter count (cubic P3)** — fixed, logs the current in-flight count.
- **Exporter knob not wired in deployments (cubic P3, coderabbit)** — valid. Added `database.statementTimeout` to the Helm chart: core configmap (`POSTGRESQL_STATEMENT_TIMEOUT`) and exporter env (`HARBOR_DATABASE_STATEMENT_TIMEOUT`), plus values.schema.json, README, and updated helm unittest snapshots. The exporter reading `HARBOR_DATABASE_*` is the same pattern as every other exporter DB knob. Compose intentionally untouched — it wires no other pool knobs and the defaults already agree.

PR #774, opened 2026-08-31
feat: Add configurable default project for project-less pushes

bupd:
## Summary

Harbor rejects registry paths without a project segment — `podman push registry.example.com/alpine:latest` fails with `invalid repository name` while the same push works against ghcr.io and Docker Hub. Docker Hub only works because the Docker CLI rewrites bare names to `library/<name>` client-side, and only for docker.io.

This adds the rewrite server-side so every OCI client (podman, oras, crane, containerd, docker) gets a flat namespace:

- **artifactinfo middleware**: bare repository paths (and blob-mount `?from=` sources) are qualified with the configured default project before routing, authorization, and the registry proxy see them — one rewrite point, everything downstream sees an ordinary `library/alpine` repo.
- **token service**: bare `repository:<name>:<actions>` scopes are qualified the same way, because CLI clients compute the token scope from the repository reference, not from the challenge.
- **config**: new `default_project_name` item (default `library`, already seeded by the initial migration, exposed via the configurations API). Empty disables the fallback and restores the previous rejection. If the configured project does not exist, the request fails at authorization like any push to a nonexistent project.

Design doc with flow diagrams and the rejected `_`-as-visible-name alternative lives in the internal task notes (05-default-project-design).

## Test plan

- Unit tests: middleware rewrite (manifest, blob upload, blob mount, qualified-path untouched, disabled mode) and token scope rewrite.
- `golangci-lint`: 0 issues; full `go build ./...` green.
- e2e for bare push/pull follows separately on the dev overlay.

## Release Notes

Pushes and pulls without a project segment (e.g. `podman push registry.example.com/alpine:latest`) now land in a configurable default project instead of being rejected, giving Harbor a Docker Hub / ghcr.io-style flat namespace for every OCI client. The default project is `library`; set `default_project_name` to another project or to empty (which disables the fallback) via the configurations API.

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -9412,6 +9412,9 @@ definitions:
       disabled_audit_log_event_types:
         $ref: '#/definitions/StringConfigItem'
         description: The audit log event types to skip to log in database
+      default_project_name:
+        $ref: '#/definitions/StringConfigItem'
+        description: The project that catches repositories pushed or pulled without a project segment. Empty disables the fallback.
   Configurations:
     type: object
     properties:
@@ -9708,6 +9711,11 @@ definitions:
       disabled_audit_log_event_types:
         type: string
         description: the list to disable log audit event types.
+        x-omitempty: true
+        x-isnullable: true
+      default_project_name:
+        type: string
+        description: The project that catches repositories pushed or pulled without a project segment. Empty disables the fallback.
         x-omitempty: true
         x-isnullable: true
   StringConfigItem:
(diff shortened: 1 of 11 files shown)


bupd:
Addressed all three cubic findings:

- **default_project_name validation**: `verifyDefaultProjectNameCfg` in the config controller now rejects values that are not legal project names (new exported `project.IsValidName`, reusing the existing regex/length rules; empty still allowed as the off switch). Unit-tested.
- **EnvKey dropped**: `DEFAULT_PROJECT_NAME` was a silent no-op for a UserScope item — removed.
- **Test singleton hygiene**: both tests now restore `default_project_name` in the shared inmemory manager (not just `DefaultCfgManager`) in their defers.

PR #780, opened 2026-08-31
feat: Add multi-format and multi-project reference schema

bupd:
Moves all schema owned by the commercial patch branches into the authoritative `harbor_next.sql`, per the no-schema-in-patches policy:

- **multi_format_* tables** (formerly numbered migration 0191, carried by `0007-multi-format-artifacts`)
- **multi_project_reference** table for the multi-project (project of projects) feature merged in 8gcr#357: FK both sides to `project`, unique (multi, sub) pair, no self-reference, rank + indexes
- the idempotent **dense-rank renormalization** (ranks are positions 1..N per multi-project, maintained by the application; this compacts pre-invariant or drifted rows on reconciliation)

Everything is `IF NOT EXISTS`/idempotent — applied by the authoritative schema reconciliation, no numbered migration. The corresponding removal of these blocks from the `0007-multi-format-artifacts` patch lands alongside: after both, patches carry code only and schema lives here.

## Release Notes

The multi-format artifact tables and the multi-project reference schema are now part of the authoritative Harbor Next schema.

### DIFF
--- a/make/migrations/postgresql/harbor_next.sql
+++ b/make/migrations/postgresql/harbor_next.sql
@@ -62,3 +62,67 @@ CREATE INDEX IF NOT EXISTS idx_claim_rules_lookup
 
 CREATE INDEX IF NOT EXISTS idx_identity_providers_jwks_cache
     ON identity_providers (id, jwks_expires_at, jwks_last_fetch_attempt);
+
+-- Multi-format artifact repositories (npm, Maven): rebuildable Postgres
+-- projection over the OCI `_index` control artifact. Authoritative mutable
+-- state lives in OCI annotations; these tables are a derived view
+-- (reconcilable from `_index`). Formerly numbered migration 0191.
+CREATE TABLE IF NOT EXISTS multi_format_package (
+  id BIGSERIAL PRIMARY KEY,
+  project_id BIGINT NOT NULL,
+  format VARCHAR(32) NOT NULL,
+  native_name VARCHAR(512) NOT NULL,
+  proj_version BIGINT NOT NULL DEFAULT 0,
+  mutable_state JSONB NOT NULL DEFAULT '{}'::jsonb,
+  last_index_digest VARCHAR(255) NOT NULL DEFAULT '',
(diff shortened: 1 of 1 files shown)


bupd:
Deep review, static + empirical against a live dev DB. **No blockers — LGTM** (posting as comment; author can't self-approve).

**Verified empirically:**
- Applied the full file twice against a live registry DB: zero errors, second run is a complete no-op (all `IF NOT EXISTS` skip, renormalization `UPDATE 0`). Idempotency holds end to end, including the formerly-0191 case where the tables already exist.
- Renormalization tested against synthetic drift (duplicate ranks, rank 0, sparse 42): heals to dense 1..N per parent, ties broken deterministically by id, second pass changes zero rows — a true fixed point. (Ran in a rolled-back transaction.)
- Constraints materialize exactly as written; the BIGINT→integer FK to `project(project_id)` is valid Postgres.

**Design fit:** one edge table with rank as a relationship attribute is the right normalization level for the flows — ranked listing and pull resolution hit `(multi_project_id, rank)`, transitive-membership lookups hit `(sub_project_id)`, the `(multi, sub)` unique key is the natural identity. No over-normalization; future per-reference policy overrides are additive columns on this table.

**Advisory (non-blocking):**
1. `REFERENCES project` without `ON DELETE` is fine **because Harbor soft-deletes projects** (row stays, `deleted=true`) and `DeleteByProject` cleans both sides at the app layer. If hard-deletion ever arrives, these FKs go RESTRICT — worth a comment line; CASCADE would then need an idempotent `ALTER` since existing DBs already hold the plain FK.
2. `rank BIGINT NOT NULL DEFAULT 0`: the app never writes 0, and renormalization sorts a stray 0 to position **1** (first). If "unset" should mean "last", drop the default; as-is it's only reachable by hand-written SQL.
3. App-side `DeleteByProject` (project deletion) removes rows without re-densifying surviving parents' ranks — ordering stays correct, but the dense invariant drifts until the next reconciliation heals it. Documented ("or drifted"); an app-side renorm there would close the window.

PR #795, opened 2026-09-07
ci: Trigger Upstream Cherry-Pick On Main Pushes

bupd:
Replace the twice-daily cron with an event-driven trigger: upstream cherry-pick candidates are re-evaluated whenever a commit lands on `main` (the `upstream-cherry-pick` concurrency group coalesces bursts — one running + one queued, extra pushes fold into the queued run). Manual `workflow_dispatch` stays for dry runs and on-demand sweeps. The fork guard now suppresses the canonical `push` trigger instead of `schedule`.


bupd:
unintended

PR #796, opened 2026-09-07
ci: Auto-bump chart appVersion to follow Harbor releases

bupd:
Adds a `chart-app-version` job to `release-please.yml` that keeps the Helm chart following Harbor app releases, plus small chart-workflow hygiene.

## What

- **New `task helm:bump-app-version NEW_APP_VERSION=vX.Y.Z`** — forward-only guard (`sort -V`): no-ops when the released version is not strictly ahead of the `appVersion` on `main`. Bumps `Chart.yaml`, regenerates `README.md` (pinned helm-docs release binary — not a source build) and helm-unittest snapshots, so the bump PR passes Chart CI as-is.
- **New `chart-app-version` job** in `release-please.yml`: runs after `publish-images` (release images exist in the registry before the chart points at them), checks out `main` regardless of the release branch, runs the task, and opens/updates a PR. One canonical branch (`chart-appversion-bump`), force-pushed — a newer release replaces a stale open bump PR instead of stacking PRs whose late merge could rewind `appVersion`. Commit type maps to the chart release line: `fix:` for a patch-only move, `feat:` when major.minor advances.
- Chart releases stay a separate cycle: the merged bump flows through the chart's own release-please (`release-please-chart`) as usual.

### Forward-only semantics

| main chart appVersion | new release | result |
|---|---|---|
| v2.15.0 | v2.15.1 | bump PR (`fix:`) |
| v2.15.0 | v2.16.0 | bump PR (`feat:`) |
| v2.16.3 | v2.15.9 (maintenance) | no-op, no PR |
| v2.16.0 | v2.16.0 | no-op |

### Hygiene (found while auditing chart workflows)

- `chart-ci.yml`: add `concurrency` (cancel superseded PR runs, matching lint-go/pr-ci) and `timeout-minutes: 20`.
- `taskfile/helm.yml`: helm-docs install hint now points at the pinned release binary; the previous `go install @latest` hint produces builds without the version footer, which fails Chart CI.

## Verified

- `task helm:bump-app-version NEW_APP_VERSION=v2.14.0` → no-op, zero diff (README/snapshot regen byte-identical with pinned helm-docs 1.14.2).
- `NEW_APP_VERSION=v2.15.1` → Chart.yaml + README badge + 106 snapshots updated, 249 unittest pass.
- actionlint clean (only its stale builtin runner-label list complains about `ubuntu-26.04`, same as every existing workflow).

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -304,3 +304,167 @@ jobs:
       component: chart
     secrets:
       SYNC_APP_PRIVATE_KEY: ${{ secrets.SYNC_APP_PRIVATE_KEY }}
+
+  chart-app-version:
+    name: Bump Chart appVersion
+    # PRs a forward-only Chart.yaml appVersion bump to main after the
+    # release images are published; maintenance releases behind main's
+    # appVersion are a no-op. The merged PR feeds release-please-chart.
+    needs: [release-please, publish-images]
+    if: ${{ needs.release-please.outputs.release_created == 'true' }}
+    runs-on: ubuntu-26.04
+    timeout-minutes: 15
+    # Serialized: overlapping main + release-X.Y releases share one bump branch.
+    concurrency:
+      group: chart-app-version-bump
+      cancel-in-progress: false
+    permissions:
+      contents: write
+      pull-requests: write
+    env:
+      NEW_APP_VERSION: ${{ needs.release-please.outputs.tag_name }}
+    steps:
+      # Always main, even for release-X.Y releases: the bump PR targets
+      # main and the guard compares against main's Chart.yaml.
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: main
+          persist-credentials: false
+
+      - name: Configure ephemeral Git authentication
+        run: |
+          askpass="${RUNNER_TEMP}/git-askpass"
+          cat > "${askpass}" <<'EOF'
+          #!/usr/bin/env bash
+          case "$1" in
+            *Username*) printf '%s\n' 'x-access-token' ;;
+            *Password*) printf '%s\n' "${GH_TOKEN}" ;;
+            *) exit 1 ;;
+          esac
+          EOF
+          chmod 700 "${askpass}"
+          echo "GIT_ASKPASS=${askpass}" >> "${GITHUB_ENV}"
+          echo "GIT_TERMINAL_PROMPT=0" >> "${GITHUB_ENV}"
+
+      - name: Load versions
+        run: |
+          grep -E '^[A-Z_]+=' versions.env >> "$GITHUB_ENV"
+          {
+            echo "HELM_CACHE_HOME=${RUNNER_TEMP}/helm/cache"
+            echo "HELM_CONFIG_HOME=${RUNNER_TEMP}/helm/config"
+            echo "HELM_DATA_HOME=${RUNNER_TEMP}/helm/data"
+          } >> "$GITHUB_ENV"
+
+      - name: Install Task
+        uses: ./.github/actions/setup-task
+
+      - name: Install Helm
+        uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
+        with:
+          version: ${{ env.HELM_VERSION }}
+
+      - name: Install helm-unittest plugin
+        # --verify=false: Helm 4 verifies plugin provenance by default;
+        # git-sourced plugins don't ship one.
+        run: helm plugin install https://github.com/helm-unittest/helm-unittest.git --version "${HELM_UNITTEST_VERSION}" --verify=false
+
+      - name: Install helm-docs
+        # Pinned release binary, NOT `go install`: source builds drop the
+        # version footer and Chart CI fails on the README diff.
+        run: |
+          arch="$(uname -m)"
+          case "${arch}" in
+            x86_64) asset_arch="x86_64" ;;
+            aarch64|arm64) asset_arch="arm64" ;;
+            *) echo "unsupported arch: ${arch}" >&2; exit 1 ;;
+          esac
+          mkdir -p "${RUNNER_TEMP}/bin"
+          curl -fsSL "https://github.com/norwoodj/helm-docs/releases/download/v${HELM_DOCS_VERSION}/helm-docs_${HELM_DOCS_VERSION}_Linux_${asset_arch}.tar.gz" \
+            | tar -xz -C "${RUNNER_TEMP}/bin" helm-docs
+          echo "${RUNNER_TEMP}/bin" >> "$GITHUB_PATH"
+
+      - name: Read current appVersion
+        # Same parser as the taskfile (helm show chart). Also checks the
+        # pending bump branch so a maintenance release cannot rewrite an
+        # open higher-version bump PR — forward-only against BOTH main
+        # and the branch.
+        id: current
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          old=$(helm show chart deploy/chart | sed -n 's/^appVersion: //p')
+          test -n "${old}"
+          echo "OLD_APP_VERSION=${old}" >> "$GITHUB_ENV"
+          pending=""
+          if git fetch --depth=1 origin chart-appversion-bump; then
+            pending=$(git show FETCH_HEAD:deploy/chart/Chart.yaml \
+              | sed -n 's/^appVersion: "\(.*\)"$/\1/p')
+          fi
+          cur=$(printf '%s\n%s\n' "${old}" "${pending}" | sort -V | tail -n1)
+          if [ "${NEW_APP_VERSION}" = "${cur}" ] \
+            || [ "$(printf '%s\n%s\n' "${cur}" "${NEW_APP_VERSION}" | sort -V | tail -n1)" != "${NEW_APP_VERSION}" ]; then
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            echo "${NEW_APP_VERSION} not ahead of main (${old}) / pending bump (${pending:-none}); skipping"
+          fi
+
+      - name: Bump appVersion (forward-only)
+        if: ${{ steps.current.outputs.skip != 'true' }}
+        run: task helm:bump-app-version NEW_APP_VERSION="${NEW_APP_VERSION}"
+
+      - name: Generate GitHub App token
+        # App-minted token so the bump PR triggers Chart CI (GITHUB_TOKEN
+        # events start no workflows). If the app cannot mint for this
+        # repo, fall back to GITHUB_TOKEN: the PR still opens, but its
+        # checks need a manual "Approve and run" (as with release-please
+        # PRs today).
+        if: ${{ steps.current.outputs.skip != 'true' }}
+        id: app-token
+        continue-on-error: true
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ vars.SYNC_APP_ID }}
+          private-key: ${{ secrets.SYNC_APP_PRIVATE_KEY }}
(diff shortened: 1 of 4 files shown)


.github/workflows/release-please.yml
```diff
@@ -252,3 +252,113 @@ jobs:
...
+          # One force-pushed branch: newer releases replace a stale bump
+          # PR instead of stacking PRs that could rewind appVersion.
+          git push --force origin "HEAD:refs/heads/${branch}"
+
+          body="Sets the chart appVersion (default image tags) to Harbor ${NEW_APP_VERSION}, released and published by the app release pipeline. README and unittest snapshots regenerated via \`task helm:bump-app-version\`.
+
+          ## Release Notes
+
+          The chart's default Harbor image tags now follow release ${NEW_APP_VERSION}."
+          gh pr create --base main --head "${branch}" \
```
bupd, line 468: **Bump PR gets zero CI.** Push + `gh pr create` run with `${{ github.token }}`, and GitHub suppresses workflow triggers for events created by `GITHUB_TOKEN` — chart-ci (`on: pull_request`, paths `deploy/chart/**`) and pr-title never run on the generated PR. If Chart Quality Gate is required, the PR sits on "Expected" forever; if not, the regenerated README/snapshots merge unverified — either way defeats "passes Chart CI as-is". Also: `gh pr create` 403s when org/repo setting "Allow GitHub Actions to create and approve pull requests" is off, and the `|| gh pr edit` fallback then fails too (no PR exists).

Repo already has the fix pattern: `actions/create-github-app-token` + `SYNC_APP_PRIVATE_KEY` (release-notes-engine.yml:89-93). Mint an app token here and use it for the push + PR creation.

.github/workflows/release-please.yml
```diff
@@ -252,3 +252,113 @@ jobs:
...
+
+          branch="chart-appversion-bump"
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git switch -c "${branch}"
+          git add deploy/chart
+          git commit -s -m "${title}"
+          # One force-pushed branch: newer releases replace a stale bump
+          # PR instead of stacking PRs that could rewind appVersion.
+          git push --force origin "HEAD:refs/heads/${branch}"
```
bupd, line 461: **Cross-branch rewind race — the exact thing this comment claims to prevent.** The forward-only guard compares NEW_APP_VERSION against *merged main* only, but this shared branch is force-pushed by every release line (job `if` only checks `release_created`, and it runs for `release-X.Y` too).

Sequence: main releases v2.16.0 → bump PR open, unmerged (main Chart.yaml still v2.15.x). release-2.15 then ships v2.15.8 → job checks out main, v2.15.8 > v2.15.x passes the guard, force-push + `gh pr edit` rewrite the open PR down to v2.15.8. Merge it and the v2.16.0 bump is silently lost until the next app release.

Guard against max(main, existing `chart-appversion-bump` branch): fetch the branch if it exists and include its appVersion in the `sort -V` comparison.

.github/workflows/release-please.yml
```diff
@@ -252,3 +252,113 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          if git diff --quiet -- deploy/chart; then
+            echo "appVersion already at or ahead of ${NEW_APP_VERSION}; no PR"
+            exit 0
+          fi
+
+          # feat when the release moves major.minor (chart minor bump via
+          # release-please), fix for a patch-only move (chart patch bump).
+          old=$(git show HEAD:deploy/chart/Chart.yaml \
```
bupd: **Second, divergent parser for the same value → silent `feat` inflation.** The Taskfile reads the old appVersion via `helm show chart | sed` (YAML-aware); this step re-parses the raw file with an optional-double-quote regex. Any formatting helm tolerates but this regex does not (single quotes, trailing comment, odd spacing) makes `old` empty or garbled, `"${old%.*}" = "${NEW_APP_VERSION%.*}"` never matches, and every patch release gets typed `feat:` — inflating the chart minor via release-please-chart with no error anywhere.

Single extraction point instead: capture the old version with the same `helm show chart` command before running the bump task (tree is still clean there), or have the task emit it.

.github/workflows/release-please.yml
```diff
@@ -252,3 +252,113 @@ jobs:
       tag_name: ${{ needs.release-please-chart.outputs.tag_name }}
     secrets:
       REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
+
+  chart-app-version:
```
bupd, line 308: **No `concurrency` group, but the job writes a branch shared across runs from different refs.** Overlapping releases (main v2.16.0 + release-2.15 v2.15.8 — plausible, since this job starts only after the long publish-images build) can interleave: run A pushes @v2.16.0, run B force-pushes @v2.15.8, run A then creates/edits the PR titled "feat: … v2.16.0" against a branch whose diff sets v2.15.8. Squash-merging that yields a feat-typed commit whose actual change is a patch (or backwards) move.

Add `concurrency: { group: chart-appversion-bump }` (no cancel) to serialize the branch writers.

taskfile/helm.yml
```diff
@@ -295,6 +297,25 @@ tasks:
     desc: "Run helm unittest and update snapshots"
     cmd: helm unittest {{.CHART_DIR}} --update-snapshot
 
+  bump-app-version:
+    desc: "Bump chart appVersion forward-only, regenerate README + snapshots (usage: task helm:bump-app-version NEW_APP_VERSION=v2.16.0)"
+    requires:
+      vars: [NEW_APP_VERSION]
+    deps: [deps:build]
+    vars:
```
bupd, line 300: **`deps:` runs before `status:` is evaluated** (verified on task 3.53.1: `NEW_APP_VERSION=v2.14.0` still executes `helm dependency build`, pulling valkey from ghcr.io before reporting "up to date"). So every guaranteed no-op — i.e. every maintenance release — still hard-depends on ghcr.io, and in the workflow the no-op path has also already installed Task/Helm/helm-unittest/helm-docs (4 external fetches) before this guard runs. A ghcr or github-releases hiccup paints the release run red for a bump that was never going to happen.

Fix here: move `deps:build` from `deps:` to the first entry of `cmds:` (cmds are skipped when status says up-to-date; do-work ordering unchanged). Optionally also gate the workflow with a cheap sed+`sort -V` check on Chart.yaml right after checkout, before any tool installs.

.github/workflows/release-please.yml
```diff
@@ -252,3 +252,113 @@ jobs:
...
+        # --verify=false: Helm 4 verifies plugin provenance by default;
+        # git-sourced plugins don't ship one.
+        run: helm plugin install https://github.com/helm-unittest/helm-unittest.git --version "${HELM_UNITTEST_VERSION}" --verify=false
+
+      - name: Install helm-docs
+        # Pinned release binary, NOT `go install`: source builds drop the
+        # version footer and Chart CI fails on the README diff.
+        run: |
+          mkdir -p "${RUNNER_TEMP}/bin"
+          curl -fsSL "https://github.com/norwoodj/helm-docs/releases/download/v${HELM_DOCS_VERSION}/helm-docs_${HELM_DOCS_VERSION}_Linux_x86_64.tar.gz" \
```
bupd: **Hardcoded `Linux_x86_64` drops the arch detection this step otherwise mirrors.** chart-ci.yml installs helm-docs behind a `uname -m` case (x86_64/arm64, added for self-hosted portability); this copy pins x86_64. Fine on GitHub-hosted ubuntu-26.04 today, but the next runner migration (repo has had several) makes the binary fail to exec — at release time, on the release-critical path.

This is the third near-verbatim copy of the Load-versions + setup-task + setup-helm + helm-unittest + helm-docs stack (chart-ci.yml, partially publish-chart.yml). Consider a composite action next to `.github/actions/setup-task` so the arch case lives in one place.

.github/workflows/release-please.yml
```diff
@@ -252,3 +252,113 @@ jobs:
...
+            | tar -xz -C "${RUNNER_TEMP}/bin" helm-docs
+          echo "${RUNNER_TEMP}/bin" >> "$GITHUB_PATH"
+
+      - name: Bump appVersion (forward-only)
+        run: task helm:bump-app-version NEW_APP_VERSION="${NEW_APP_VERSION}"
+
+      - name: Configure ephemeral Git authentication
```
bupd: **Third verbatim copy of this askpass block in this one file** (release-please job line 58, create-maintenance-branch line 171, now here; a variant also lives in `.github/scripts/update-release-notes.sh`). The repo already uses local composite actions (`.github/actions/setup-task`, `setup-go-cached`) — extract e.g. `.github/actions/git-askpass` and reuse. (If the bump PR moves to an app token per the other comment, this copy disappears entirely: checkout can persist that token.) The `git config` bot-identity pair below is similarly re-typed from backport.yml and belongs in the same action.

taskfile/helm.yml
```diff
@@ -252,9 +252,10 @@ tasks:
     preconditions:
       - sh: command -v helm-docs
         msg: |
-          helm-docs not found. Install:
+          helm-docs not found. Use the release binary, not `go install`
```
bupd: **Warns in prose (twice) what it could enforce once.** This PR had to edit the identical message in two tasks (`helm-docs` here and `_require-helm-docs` below) — direct evidence of the duplication cost; `docs` already consolidates via `deps: [_require-helm-docs]`, this task could too. Deeper: `command -v helm-docs` still passes for the exact go-install build the message warns about. The pin is already in scope — enforce it:

```yaml
- sh: helm-docs --version | grep -qF "{{.HELM_DOCS_VERSION}}"
```

Source builds report a different/dev version, so the footer failure mode gets rejected mechanically instead of documented in two (now three, counting the workflow comment) places.

bupd:
Addressed 7 of the 8 review findings in the latest push (a07679fd):

- **App token for the bump PR**: `create-github-app-token` (SYNC_APP) with `continue-on-error` + documented `GITHUB_TOKEN` fallback, so bump-PR checks trigger automatically when the app can mint for this repo.
- **Pending-PR rewind**: the guard now fetches the `chart-appversion-bump` branch and compares against max(main, pending), so a maintenance release can no longer rewrite an open higher-version bump PR.
- **Single parser**: the workflow reads the old appVersion via `helm show chart` (same as the taskfile) before bumping; the raw `git show`+sed parse is gone.
- **Concurrency**: job serialized via `concurrency: chart-app-version-bump` (no cancel).
- **No-op network use**: `deps:build` moved from `deps:` into `cmds:` after the `status:` guard — skipped bumps touch nothing.
- **helm-docs arch**: install step now carries the same uname arch case as chart-ci.
- **Pinned-binary enforcement**: `_require-helm-docs` now greps `helm-docs --version` for the versions.env pin (a `go install` build fails the precondition), and the check task reuses it instead of a duplicate block.

Deferred: askpass-block dedup into a composite action — third copy exists pre-PR in this file; extracting all of them is a separate cleanup.

bupd:
Latest push (63ff6c15) addresses the new review round and stacks this PR on #814:

**Fixed (cubic):**
- Empty/missing `appVersion` now fails fast: new precondition rejects an empty `CURRENT` before `status`/`sed` can silently no-op.
- helm-docs pin check is now an exact version compare (`awk $3` equality), so v1.14.20 no longer satisfies a 1.14.2 pin.

**Already fixed in an earlier push (coderabbit, stale anchors):** job-level `concurrency: chart-app-version-bump` + the pending-branch fetch-and-compare guard landed in a07679fd — both suggested remedies are in place.

**Declined with reasons:**
- curl/tar/gh setup steps: guaranteed on GitHub-hosted ubuntu runners and used bare throughout every existing workflow in this repo; a self-hosted bootstrap story belongs in a shared composite action, not per-job.
- `Install envsubst` in publish-chart.yml: intentionally kept (maintainer decision) for a possible return to self-hosted runners; commented as such in the workflow.
- zizmor `$/` self-repository syntax: the repo consistently uses `./.github/actions/setup-task` everywhere; switching one call site would be inconsistent — a repo-wide sweep can adopt it separately.

Also rebased into the stack: base is now #814 (`next-chart-relnotes`), which sits on #813 (`next-chart-rename`).

PR #814, opened 2026-09-07
ci: Render chart release notes with the release-notes engine

bupd:
Chart releases currently ship release-please's generic changelog. This gives them the same treatment as app releases: **Summary** (verbatim `## Release Notes` sections from the merged PRs, `(#N)` references only — nothing synthesized), attributed **What's Changed**, an install snippet, and cosign verification for the chart artifact.

## Design: chart mode in the existing engine, not a second pipeline

`update-release-notes.sh` already does 90% of the work; the What's-Changed content comes from the release-please CHANGELOG, and the chart line has its own `deploy/chart/CHANGELOG.md` that is *already scoped to chart commits* — so path-scoping came for free. Changes:

- **`update-release-notes.sh`**: accepts `chart-vX.Y.Z` tags → chart mode. Sources `deploy/chart/CHANGELOG.md`; passes `previous_tag_name=<previous chart-v* tag>` to `generate-notes` so authorship is computed between *chart* tags, never against the latest app release. Skips app-only blocks (commercial patches, container-image table, New Contributors — the latter would credit unrelated app PRs in a chart tag range). Adds a chart cosign-verify block (`publish-chart.yml@refs/heads/main` identity) and a "default image tags follow appVersion" footer. The chart name is read from `Chart.yaml` at the tag — the upcoming harbor-next→harbor rename needs **no change** here, and the app-release "Helm Chart" snippet now uses the same dynamic name instead of a hardcoded `harbor-next`.
- **`extract-changelog-release.mjs`**: also matches release-please's unbracketed first-release heading style (`## 2.0.0 (date)`), which the chart changelog uses.
- **`release-notes-engine.yml`**: new optional `component` input (`''` = app, `chart`), plumbed as `RELEASE_NOTES_COMPONENT`. App callers unchanged.
- **`release-please.yml`**: `update-chart-release-notes` job runs the engine after `chart` publish (install snippet points at a pullable artifact); `preview-chart-release-notes` mirrors the app preview on the chart release PR (reads the bumped `.release-please-manifest-chart.json` on the PR branch). `release-please-chart` exposes `prs_created`/`prs` (top-level action outputs, not path-prefixed).

## Verified

Full dry run in chart mode against the chart 2.0.0 changelog + PR #56: Summary renders #56's Release Notes section verbatim, What's Changed attributes @Vad1mo via generate-notes, install/verify snippets carry the right chart name and version. `bash -n` + `node --check` + actionlint clean (its stale builtin runner list flags `ubuntu-26.04` as always; the "component input not defined" it reports when run from the repo root is a false positive from resolving the reusable workflow in a different worktree).

## Composes with

- The rename PR (harbor-next → harbor): nothing here hardcodes the chart name.
- The release-as unstick + PR 796 (appVersion bump): independent files except `release-please.yml` job additions at the end — trivial rebase if they land first.

### DIFF
--- a/.github/scripts/extract-changelog-release.mjs
+++ b/.github/scripts/extract-changelog-release.mjs
@@ -7,7 +7,9 @@ if (!changelogPath || !version || !outputPath) {
 }
 
 const lines = readFileSync(changelogPath, 'utf8').split(/\r?\n/);
-const headings = [`## [${version}]`, `## [v${version}]`];
+// release-please writes "## [x.y.z](compare-url) (date)" when a previous tag
+// exists and "## x.y.z (date)" for a line's first release (e.g. the chart).
+const headings = [`## [${version}]`, `## [v${version}]`, `## ${version} `, `## v${version} `];
 const start = lines.findIndex(line => headings.some(heading => line.startsWith(heading)));
 
 if (start === -1) {
--- a/.github/scripts/update-release-notes.sh
+++ b/.github/scripts/update-release-notes.sh
@@ -18,7 +18,12 @@ github_retry() {
 }
 
 preview_pr_number="${RELEASE_NOTES_PREVIEW_PR_NUMBER:-}"
-if [[ -n "${preview_pr_number}" && ! "${TAG_NAME:-}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+component="${RELEASE_NOTES_COMPONENT:-}"
+if [[ -n "${preview_pr_number}" && "${component}" == "chart" ]]; then
+  # A chart release PR bumps the chart manifest on its branch to the new version.
+  preview_version=$(node -e "const manifest = require('./.release-please-manifest-chart.json'); const version = manifest['deploy/chart']; if (!version) { throw new Error('missing chart release version'); } console.log(version);")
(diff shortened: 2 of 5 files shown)


bupd:
Review disposition (pushed in 0b9b9fac):

**Fixed:**
- Predecessor tag selection: now picks the greatest `chart-v*` tag strictly *below* the target (descending sort, line after target), so recreating an older chart release no longer diffs against a newer tag; previews keep the latest-tag fallback.
- Bootstrap chart release: with no `chart-v*` tag at all, the generate-notes call is skipped (empty What's Changed) instead of letting GitHub diff against the latest app release and attribute unrelated app PRs. format-release-notes.mjs handles the empty input (pure regex scans).

**Declined:**
- Extracting `chart_mode` into a separate module: the mode flag was a deliberate choice to reuse one pipeline (~6 branch points) rather than fork it; extraction doubles the surface for a second consumer that shares 90% of the logic. Revisit if a third mode appears.
- Test harness for extract-changelog-release.mjs: the repo has no script test infra to hook into; the unbracketed-heading path was verified against the real chart CHANGELOG in a dry run. A scripts test harness is worthwhile but is its own change.
- zizmor `$/` self-repository notes: repo-wide `./` convention, separate sweep.

PR #817, opened 2026-09-07
chore: Sync Patch Release Changelog Sections To Main Automatically

bupd:
Companion to the 2.15.x backfill PR. Patch releases cut from `release-X.Y` update only that branch's `CHANGELOG.md`, so main's changelog silently stops at `X.Y.0`.

Adds a `sync-main-changelog` job to `release-please.yml`: when a maintenance release publishes, it extracts the released section (existing `extract-changelog-release.mjs`), inserts it into main's `CHANGELOG.md` in version-sorted position via the new `insert-changelog-section.mjs` (idempotent — re-runs and already-present versions no-op; a 2.15.9 section lands below an existing 2.16.0), and opens a `chore:` PR against main for a maintainer to merge.

**Cherry-pick note:** the workflow runs from the *release branch's* copy on patch releases, so this needs to be cherry-picked onto `release-2.15` (before v2.15.9) to take effect there — `chore:` type keeps that cherry-pick from bumping a release.

Known limitation: the sync PR is created with `GITHUB_TOKEN`, so it won't trigger PR CI automatically (changelog-only change; close/reopen re-triggers if checks are required).

### DIFF
--- a/.github/scripts/insert-changelog-section.mjs
+++ b/.github/scripts/insert-changelog-section.mjs
@@ -1,0 +1,56 @@
+import { readFileSync, writeFileSync } from 'node:fs';
+
+const [changelogPath, sectionPath, version] = process.argv.slice(2);
+
+if (!changelogPath || !sectionPath || !version) {
+  throw new Error('usage: insert-changelog-section.mjs <changelog> <section> <version>');
+}
+
+if (!/^\d+\.\d+\.\d+$/.test(version)) {
+  throw new Error(`version must be X.Y.Z, got: ${version}`);
+}
+
+const section = `${readFileSync(sectionPath, 'utf8').trim()}\n`;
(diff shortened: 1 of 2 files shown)


bupd:
Review findings addressed:

- **Fixed:** legacy-block ordering — a version older than a `## vX.Y.x` block now inserts after it (major/minor compared); duplicate detection now also matches legacy nested headings (`### vX.Y.Z`); the sync branch is force-pushed and PR creation is skipped when one is already open, so a rerun after a partial failure recovers cleanly.
- **Not adopted:** extracting the ephemeral git-auth block into a shared composite action — deliberate for now: this PR must stay minimal to cherry-pick cleanly onto `release-2.15`, and the composite refactor would touch the two existing jobs too. Worth a main-only follow-up.

PR #824, opened 2026-09-08
ci: Render release PR note previews whenever the PR is open

bupd:
Fixes the pipeline gap that left #822 without its `## Summary`: the preview jobs keyed off release-please's `prs_created`, which is true only in the run that creates or updates the release PR. A release PR created before an engine change (like #814's chart mode) kept a stale preview forever — nothing in the pipeline refreshed it.

## What

- New `find-release-prs` job locates the open app + chart release PRs by their deterministic branch names (`release-please--branches--main`, `…--components--chart`).
- Both preview jobs run whenever their release PR is open (still gated on no release cut in the same run). Re-rendering is idempotent.

After this merges, the next push to main re-renders #822's preview through the pipeline itself — no manual runs.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -124,17 +124,44 @@ jobs:
           git commit -s -m "chore: target next development version ${next_version}"
           git push
 
+  find-release-prs:
+    name: Find Open Release PRs
+    # Previews key off "release PR open", not prs_created, so an
+    # untouched open PR still re-renders after engine changes. Runs after
+    # both release-please instances so a PR created this run is found.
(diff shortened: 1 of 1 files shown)


bupd:
Race fixed in 55ca5ffa: `find-release-prs` now runs **after** both release-please instances (`needs: [release-please, release-please-chart]`, `!cancelled()` so maintenance refs — where the chart instance is skipped — still work), so a PR created in the same run exists before the query. Also restored maintenance-branch previews, which the first version had accidentally made main-only: the app lookup now derives head/base from `${GITHUB_REF_NAME}` (`release-please--branches--<ref>`), chart lookup stays main-only.

PR #819, opened 2026-09-07
ci: Support Patch Stream Multi-Commit Chains In Release Pipeline

bupd:
## Summary

Companion to **container-registry/8gcr#389 (Patch Stream)** — patch branches become linear chains of squash-merged PRs on a pinned baseline; this PR teaches the harbor-next release pipeline to consume them. **Merge this BEFORE 8gcr#389's flow goes live** (everything here is backward-compatible with today's single-commit branches, so merge order relative to current operation is safe).

### Changes

- **`taskfile/release-ready.yml`** — `_validate-patches` walks each branch as a chain: every commit non-merge, root parented on the target or (for branches declared in the new `taskfile/commercial-patch-stacks`) anywhere on the declared parent's chain. `_duplicate-patches` duplicates full chains as one connected jj set (`(::tips) ~ ::target`) — stacking topology survives, single-commit branches are a chain of one.
- **`taskfile/commercial-patch-stacks`** (new) — declared stacking: `0006-aws-rds-iam-auth 0005-pgx-monitoring`. Gates verify the declared shape; nothing is inferred from ancestry.
- **`.github/scripts/update-release-notes.sh`** — commercial section becomes per-feature: chain PRs whose `(#N)` is not in the previous `patch-record/<tag>` tag's recorded PR set (rebase-proof set diff), rendered with each PR's `## Release Notes` body lines; "_No changes this release_" collapse; shallow fetches bounded by `--shallow-exclude=patch-base/<line>`. Legacy fallback (flat tip-subject list) until the stream is seeded.
- **`.github/workflows/notify-fork-sync.yml`** — fires from `main` **and** `release-X.Y`, payload carries `line` (feeds 8gcr's dry `upstream-check`; supersedes #798).
- **`.github/workflows/release-please.yml`** — new `dispatch-release-cut` job after notes render: sends `release-cut {tag, target_sha, line}` to 8gcr (bump + record). Render first, stamp second.

### Verified

Sandbox (chain + stacked + mid-chain-rooted states): chain-aware validation accepts single-commit, full-chain, and mid-chain-stacked shapes; rejects merge commits and orphan-rooted stacked branches loudly; `jj duplicate` of the connected set preserves chain order + stacking onto a fresh target with zero conflicts. `bash -n` + `actionlint` clean (only pre-existing `ubuntu-26.04` label gripes from a stale local label DB).

[...]

PR #829, opened 2026-09-08
ci: Dispatch fork sync for release branches with scoped token

bupd:
Split 1/4 of #819 (independent, mergeable now): the 8gcr upstream conflict check now also fires on `release-X.Y` pushes and carries the branch (`line`) in a jq-built, JSON-safe payload; the SYNC_APP token is scoped to `contents: write`.

### DIFF
--- a/.github/workflows/notify-fork-sync.yml
+++ b/.github/workflows/notify-fork-sync.yml
@@ -2,13 +2,16 @@ name: Notify Fork Sync
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - "release-[0-9]*.[0-9]*"
 
 permissions: {}
 
 jobs:
   dispatch:
     runs-on: ubuntu-26.04
+    timeout-minutes: 5
     steps:
       - name: Generate GitHub App token
         id: app-token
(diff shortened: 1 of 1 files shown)


bupd:
## Recommend closing — superseded by #840, salvage two lines

#840 (`ci: Dispatch Fork Sync From Release Branches With Line Payload`) merged on 2026-09-08 and
carries this PR's intent. Checked against `main` today:

```
$ git show main:.github/workflows/notify-fork-sync.yml
on:
  push:
    branches: [main, "release-[0-9]*.[0-9]*"]     # release lines: already there
...
          payload=$(jq -n --arg sha "${SHA}" --arg line "${LINE}" \
            '{event_type: "upstream-sync", client_payload: {sha: $sha, line: $line}}')
```

Release-branch triggering and the jq-built `line` payload are both live. What is left in this
branch after that is:

| Residue | Verdict |
|---|---|
| `permission-contents: write` on the app token | **worth keeping** — without it the token inherits every permission of the 8gcr installation (zizmor `github-app`) |
| `timeout-minutes: 5` on the job | worth keeping, one line |
| `jq -cn ... \| curl --data-binary @-` instead of a shell variable | equivalent; no behaviour change |
| YAML list reflow, step rename, comment rewording | cosmetic |

Two of those are real hardening and the rest is noise, so rebasing this branch is not the cheap
path — the conflict is against the very commit that superseded it. A fresh three-line PR adding
`permission-contents: write` and `timeout-minutes: 5` to main's current file gets the value without
the merge.

Recommending closure. Not closing it — that call is yours.

bupd:
Closing as superseded by merged #840. The two lines worth keeping — `permission-contents: write` on the app token and `timeout-minutes: 5` — are not on main and will be re-filed as a small standalone PR against the current `notify-fork-sync.yml`.

PR #830, opened 2026-09-08
ci: Consume Patch Stream chains in release notes and dispatch release cuts

bupd:
Split 2/4 of #819 (core; backward-compatible — falls back to today's flat tip-subject list until 8gcr seeds `patch-base/*`): release notes compute each release's commercial delta as chain PRs absent from the previous `patch-record` tag (rebase-proof), and a post-notes `dispatch-release-cut` job tells 8gcr to bump baselines and stamp `patch-record/<tag>`. Payloads jq-built, token contents-scoped.

Open cubic finding inherited from #819 for the author: the patch-record selector should match records by `line:` before computing the seen-PR set.

PR #831, opened 2026-09-08
ci: Validate commercial patch chains in release-ready

bupd:
Split 3/4 of #819 (compatible with current single-commit branches): release-ready validates patch branches as linear chains — root must parent on the target or on its declared parent branch (`taskfile/commercial-patch-stacks`, optional file); merge commits and undeclared stacking are rejected; apply-patches duplicates full chains.

Open cubic findings inherited from #819 for the author: traversal depth cap, stacked-root tip-equality vs 8gcr's gate, fixture tests.

### DIFF
--- a/taskfile/release-ready.yml
+++ b/taskfile/release-ready.yml
@@ -4,6 +4,7 @@ vars:
   PATCHES_REPO: '{{.PATCHES_REPO | default "github.com/container-registry/8gcr"}}'
   PATCHES_CLONE_URL: '{{.PATCHES_CLONE_URL | default "git@github.com:container-registry/8gcr.git"}}'
   PATCHES_SERIES: '{{.PATCHES_SERIES | default (printf "%s/taskfile/commercial-patches" .ROOT_DIR)}}'
+  PATCHES_STACKS: '{{.PATCHES_STACKS | default (printf "%s/taskfile/commercial-patch-stacks" .ROOT_DIR)}}'
   PATCH_AUTH_DIR: '{{.PATCH_AUTH_DIR | default ".opencode/tmp/patch-auth"}}'
   RELEASE_READY_DIR: '{{.RELEASE_READY_DIR | default ".opencode/tmp/release-ready"}}'
   PATCHES_REMOTE_URL:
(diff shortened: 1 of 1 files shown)


bupd:
Flagging this before it gets rebased: I think merging it would red-line `release-ready` for every patch branch at once.

The new `_validate-patches` walks each patch branch as a strictly linear chain and rejects merge commits:

```
if [ "$(wc -w <<< "${parents}")" -ne 1 ]; then
  echo "commercial patch chain contains a merge commit: ..."
  exit 1
```

Under the v3 merge-sync model that landed in #840–#843, **every** patch branch receives a bot sync-merge commit on every dispatch. So this validator would fail all 8 branches on its first run. `main`'s current comment on that task says as much explicitly:

> Patch branches carry real multi-commit history plus sync-merge commits, so no commit-shape rule applies.

`#841` already replaced this exact task — `_validate-patches` is now a merge-base check and `_octopus-patches` uses `jj new "heads(base|tips)"` rather than `jj duplicate` + `jj new`. That's also why this conflicts: the diff is written against the pre-#841 base.

Same applies to the rest of the set, as far as I can tell — #830's `patch-record`/`patch-base` PR-set model and dispatch job are superseded by #842/#843 (main has `mark-commercial-release` and the `changelogs/<branch>.md` reader), and #832's own body says "DO NOT MERGE until the 8gcr Patch Stream restack (8gcr#389) cutover", which was closed as superseded on 2026-09-08.

**One thing is worth salvaging before these get closed.** #829 carries two lines that are genuinely not on `main`: `permission-contents: write` and `timeout-minutes: 5` on `notify-fork-sync.yml`. Main's `create-github-app-token` step scopes `owner`/`repositories` to 8gcr but sets no `permission-*` inputs, so the minted token inherits every permission the SYNC_APP installation holds on that repo. That's a real (if repo-scoped) hardening gap, and it's a one-file PR.

This is `patchops`' territory, so I'm not touching any of it — just recording the CI hazard so it doesn't get rebased and merged on the assumption that a conflict is the only thing wrong with it.

Nothing tested here: these are `taskfile/`-only changes and I did not run `release-ready`.

bupd:
## Recommend closing — it would red-line `release-ready` for all eight patch branches

#841 merged on 2026-09-08 and rewrote the same two tasks in `taskfile/release-ready.yml` in the
opposite direction. Main's `_validate-patches` now asks only that a branch share history with the
target:

```
$ sed -n '118,131p' taskfile/release-ready.yml
  _validate-patches:
        if ! git merge-base "refs/remotes/patches/${branch}" "${base}" >/dev/null; then
```

Commit shape stopped being a constraint on purpose, because the merge-sync model gives every patch
branch real multi-commit history plus sync-merge commits. This PR reintroduces the rule that #841
removed:

```
$ git show <head>:taskfile/release-ready.yml | sed -n '155p'
              echo "commercial patch chain contains a merge commit: ${branch} at ${c}" >&2
```

Counted against the live 8gcr refs today:

| branch | commits above merge-base with `next/main` | merge commits |
|---|---|---|
| 0001-commercial-feature-gate | 5 | **4** |
| 0002-branding | 4 | **3** |
| 0003-sftp-replication | 4 | **3** |
| 0004-federated-robot-accounts | 4 | **3** |
| 0005-pgx-monitoring | 4 | **3** |
| 0006-aws-rds-iam-auth | 5 | **3** |
| 0007-multi-format-artifacts | 5 | **4** |
| 0008-audit-log-otel | 4 | **3** |

Every one of the eight fails the merge-commit check on its first commit. `task release-ready` and
`task apply-patches` go red immediately, and the `dev` overlay along with them.

The apply path conflicts too: main runs `_octopus-patches` (`jj new "heads(...)"`), this PR restores
`_duplicate-patches` (`jj duplicate`). And its declared-parent lookup reads
`taskfile/commercial-patch-stacks`, which main's pipeline does not have and does not consult.

The parent PR #819 is closed; so is 8gcr#389, the restack this validation was written for.

Recommending closure. Not closing it — that call is yours.

bupd:
Closing as superseded by merged #841, which removed the commit-shape rule this PR reinstates. All eight live patch branches carry 3–4 merge commits, so this would red-line `release-ready` on merge.

PR #833, opened 2026-09-08
fix(orm): Detach AfterCommit hooks from a cloned or copied context

Vad1mo:
## Summary

`orm.Clone` and `orm.Copy` install a fresh `orm.NewOrm()` but keep the parent's context values, so a transaction started from a copied transaction context still found the enclosing `AfterCommit` hooks sink. It committed on its own session and then handed its hooks to a scope that had possibly already committed and drained. The same context used outside any transaction of its own queued the callback into that foreign sink instead of running it inline. Either way the cache eviction was delayed or lost. This is the background-handler pattern in `pkg/notifier`, `controller/retention` and `controller/proxy`.

`NewContext` now detaches an inherited sink whenever it installs a non-transactional ormer, so a `Clone` or `Copy` carries no scope: a transaction started from it fires its own hooks after its own commit, and `AfterCommit` on it runs inline.

Found by Copilot review on goharbor/harbor#23114; this PR mirrors the upstream fix so the cherry-pick applies without conflicts.

## Tests

- `TestAfterCommit_IndependentTxFromCopiedContext` (Copy and Clone subtests) and `TestAfterCommit_CopiedContextOutsideTxRunsInline` in `lib/orm/test`. Both fail before this change and pass after.
- `TestDelete_ConcurrentInlineCleanup` in `pkg/cached/artifact/redis` now collects the `Delete` errors and asserts the exact eviction count (2 deletes × 2 keys) instead of "called at least once".
- `OrmSuite` (`-tags db`) and `pkg/cached/...` pass under `-race` against Postgres.

## Upstream sync note

After this merges, the cherry-pick of goharbor/harbor#23114 applies cleanly except for `src/pkg/cached/project_metadata/redis/manager.go`, whose remaining hunk comes from goharbor/harbor#23762 (`UpdateWithValidation` + `cleanUpAll`), already open as #731. Resolve #731 so that file matches the upstream version of #23114 and the later pick is conflict-free. Verified by simulating the squash-and-cherry-pick locally.

### DIFF
--- a/src/lib/orm/orm.go
+++ b/src/lib/orm/orm.go
@@ -85,6 +85,17 @@ func NewContext(ctx context.Context, o orm.QueryExecutor) context.Context {
 	if ctx == nil {
 		ctx = context.Background()
 	}
+	if _, isTx := o.(orm.TxOrmer); !isTx {
+		// A non-transactional ormer opens its own database session, so a hooks
+		// sink inherited through ctx (Clone, Copy) belongs to a transaction this
+		// session is not part of. Detach it, otherwise AfterCommit would queue
+		// into a scope that may already have committed and drained, and
+		// WithTransaction would hand its hooks to that scope instead of firing
+		// them after its own commit.
+		if _, inherited := ctx.Value(hooksKey{}).(*txHooks); inherited {
+			ctx = context.WithValue(ctx, hooksKey{}, (*txHooks)(nil))
+		}
+	}
 	return context.WithValue(ctx, ormKey{}, o)
 }
 
--- a/src/lib/orm/test/orm_test.go
+++ b/src/lib/orm/test/orm_test.go
@@ -569,6 +569,60 @@ func (suite *OrmSuite) TestAfterCommit_RejectsCompletedNestedContext() {
 	suite.False(secondScopeRan, "a completed transaction context must be rejected before starting DB work")
 }
 
+// TestAfterCommit_IndependentTxFromCopiedContext asserts that a transaction
+// started from a Copy or Clone of a transaction context is independent: it
+// fires its own hooks on its own commit instead of handing them to the scope
+// it was copied from. This is the background-handler pattern (notifier,
+// retention, proxy) where the enclosing scope may already have drained by the
+// time the copied context is used.
+func (suite *OrmSuite) TestAfterCommit_IndependentTxFromCopiedContext() {
+	for name, derive := range map[string]func(context.Context) context.Context{
+		"copy":  Copy,
+		"clone": Clone,
+	} {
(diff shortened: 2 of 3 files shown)


bupd:
Ran this against a real postgres rather than relying on the CI badge.

```
cd src && go test -tags db -count=1 -p 1 -v ./lib/orm/test/
ok  github.com/goharbor/harbor/src/lib/orm/test  0.786s
```

All 23 subtests pass, including the two new ones:

```
--- PASS: TestRunOrmSuite/TestAfterCommit_IndependentTxFromCopiedContext/copy
--- PASS: TestRunOrmSuite/TestAfterCommit_IndependentTxFromCopiedContext/clone
--- PASS: TestRunOrmSuite/TestAfterCommit_CopiedContextOutsideTxRunsInline
```

and the cache manager under the race detector:

```
cd src && go test -race ./pkg/cached/artifact/redis/
ok  github.com/goharbor/harbor/src/pkg/cached/artifact/redis  1.030s
```

`go build ./...` is clean. The branch is 9 commits behind `main`, which is close enough that the merge result should match what I tested.

I specifically checked whether hooks can now be silently dropped, since that would be worse than the bug being fixed. They cannot: the no-sink fallback in `AfterCommit` is `safeInvoke(fn)`, so the hook runs inline rather than being discarded. The stored value is a typed-nil `(*txHooks)(nil)` and both consumers handle it — `AfterCommit` guards with `ok && h != nil`, and `WithTransaction`'s `parentHooks != nil` check correctly treats it as the outermost scope.

## The thing worth one more test

What changes is timing, not delivery. A hook registered on a `Clone`/`Copy` taken inside an open transaction now fires **immediately** instead of after the outer commit. For cache eviction that means evicting *before* the commit lands, which leaves a window where a concurrent reader repopulates the cache from pre-commit state and the stale entry survives the commit — the inverse of the bug this fixes.

I don't think that's a reason to hold the PR; the current behavior (hook attached to a transaction it has no business being attached to) is clearly worse. But neither new test covers it, and `TestDelete_ConcurrentInlineCleanup` asserts call counts rather than the resulting cache state. A test that writes in a transaction, evicts via a copied context, reads concurrently, then commits and asserts the cache is empty would pin the behavior either way.

## Not tested

No end-to-end exercise of the eight real `orm.Clone`/`orm.Copy` call sites (`core/main.go:232`, `controller/retention`, `controller/proxy` ×2, `pkg/notifier`, `pkg/auditext/event/project`, `handler/project` ×2) — only the synthetic suite.

This looks like a good 2.15.9 backport candidate to me: one-line semantic fix, mirrors upstream goharbor/harbor#23114, no API or schema change.

Issue #839, opened 2026-09-08
ADR: Instance storage, global quota and storage usage reporting


Vad1mo:
status: accepted
date: 2026-09-08
deciders: @vad1mo

Decision log:
- 2026-09-08 proposed as a single design; restructured into two themes the same day.
- 2026-09-08 accepted. Theme A read access: system administrators only. Theme B direction: pull (core fetches from registryctl). Implementation: `feat/global-quota` (theme A) and `feat/storage-reporting` (theme B), one PR each.
format: [MADR 3.0](https://adr.github.io/madr/), as used in harbor-satellite `docs/decisions/`

## Context and Problem Statement

Operators run Harbor for tenants. Tenants see their project quota but never the capacity of the instance, so they cannot judge headroom, and operators cannot express an instance-wide limit. Harbor also has no truthful source for physical usage: it embeds docker/distribution, whose storage drivers expose no capacity API, and `GET /systeminfo/volumes` statfs's the core container's `/data`, which is not the registry volume in Kubernetes. Upstream requests goharbor/harbor#16671 and #18011 have been open since 2022 without a design.

This record covers **two independent features** that together answer the customer need but ship and work on their own:

- **Theme A, Global quota**: an instance-wide storage limit with the usual CRUD, visible to system administrators, optionally enforced. The customer's "users" are the administrators of handed-over instances, so admin visibility is the requirement; it also matches the existing admin-only "Quota used" card on `/harbor/projects` (`statistics-panel.component.ts:64-67`, `statistic.go:84-109`).
- **Theme B, Storage usage reporting**: Harbor learns the real capacity and usage of its storage volume. For now this is only possible for filesystem storage (PVC or bind mount).

A works without B (usage comes from Harbor's own accounting). B works without A (the numbers are shown as information and fix `/systeminfo/volumes`). When both are present, B's measured usage replaces A's accounted usage in the view and in enforcement.

## Decision Drivers

- Filesystem storage is the majority case and must need zero operator setup.
- No new contention on the push path; the per-project `quota_usage` row lock already serializes pushes within a project.
- Users must never be left unable to free space when a limit is reached.
- Every number shown carries its provenance and freshness.
- Additive only: no schema or API breakage, defaults preserve current behaviour.
- Each theme must be shippable and useful alone.

---

### Problem

Where does the instance-wide limit live, who can read and write it, what does "used" mean when nothing measures the backend, and how is the limit enforced?

[...]

PR #845, opened 2026-09-08
feat: Add Global Storage Quota

Vad1mo:
## Summary

Theme A of the decision record #839: an instance-wide storage quota.

- `GET/PUT/DELETE /api/v2.0/system/quota` (new swagger tag `systemquota`); read for system administrators and system robots with `system-quota:read`, writes for administrators only.
- Effective view: hard limit, used, free, `used_source` (`accounted` from Harbor's blob table, or `measured` when a measurer is installed by storage usage reporting, theme B), the advisory sum of project quotas and the number of unlimited projects.
- Opt-in enforcement: a threshold gate on manifest PUT, blob upload POST/PATCH/PUT and artifact copy. Pure read of cached state, no reservation row, no CAS. DELETE, GC and admin APIs are never blocked.
- Singleton table `system_quota` in `harbor_next.sql` (additive, idempotent).
- Portal: the admin "Quota used" card shows used of limit with a bar and the usage source; the Project Quotas page gets a global quota editor (enter `-1` to remove); robot permission label; all ten language files.
- Exporter gauges `harbor_system_quota_hard_bytes`, `harbor_system_quota_used_bytes{source}`, `harbor_system_quota_enforce` (in-core exporter mode).
- Docs: `docs/global-quota.md`.

## Related Issues

Implements #839 (theme A). Upstream context: goharbor/harbor#16671, goharbor/harbor#18011.

## Type of Change

- [x] New feature (`feat:`)

## Release Notes

Adds a **global storage quota** for the whole instance. System administrators set a hard limit through the new `GET/PUT/DELETE /api/v2.0/system/quota` API or on the Project Quotas page. The projects page then shows instance usage against that limit, with the source of the usage value (Harbor's blob accounting, or a measured value where storage usage reporting is available) and the sum of all project quotas so over-commitment is visible. Optionally the quota can be **enforced**: once usage reaches the limit, pushes are denied with a clear message while deletions and garbage collection keep working so space can always be freed. Enforcement is off by default and nothing changes for existing installations until a quota is set. See `docs/global-quota.md`.

## Testing

- [x] Unit tests added/updated (controller, DAO with `db` tag, handler, gate middleware, exporter collector, RBAC, Karma specs for the two portal components)

Local runs: `go test` for all touched packages, `golangci-lint` (0 issues), spectral API lint (0 errors), `ng lint`, and the 10 affected Karma specs. The DB-tagged DAO test runs in CI.

## Checklist


Vad1mo:
Review pass over the bot threads (CodeRabbit, Copilot, cubic). Addressed in b61697ab8 and 2975a2ada, plus the shard fix below.

**Fixed**
- `CheckCapacity` now reads one quota snapshot and the cached usage only; the project allocation aggregate is reporting-only and never on the push path. It denies once usage reaches the limit (`>=`) and compares by subtraction, so a huge declared `Content-Length` cannot overflow.
- The gate counts `Content-Length` only for blob uploads (PATCH/PUT on `/blobs/uploads/`); a manifest body is metadata and is checked against usage alone. Docs reworded accordingly.
- Project allocation joins `project` and ignores deleted projects; `ProjectAllocation` moved to the model package so the manager contract does not leak the DAO.
- `PUT /system/quota` takes a storage-only `hard` (`minimum: 1`) and a required `enforce`, and declares 422. Omitting `enforce` can no longer silently switch enforcement off.
- `SetMeasurer` logs when the controller does not accept a measurer.
- Portal: only a 404 marks the quota as unset, other errors keep the last known state; the Project Quotas page shows "not set" only after a confirmed 404 and disables the editor while unknown; typed form value; the modal keeps a single `valueChanges` subscription; Karma covers the card and the editor.
- Non-German locales keep the English strings (the portal convention cubic pointed at).
- Docs: `curl -u admin` without the password on the command line.
- DAO test creates the `system_quota` table from `harbor_next.sql` (the harness runs only numbered migrations), asserts exact deltas against projects it creates itself, and covers a deleted project.
- `pkg/tag/dao` FK test used artifact id `2`, which collides with the suite's own artifact once a new db-tagged package shifts the shard split (ids come from a sequence shared by all DB test packages). It now uses an impossible id.

**Not changed, with reasons**
- Chunked uploads without `Content-Length` and multi-chunk uploads: checked against usage alone. Rejecting unknown-length bodies would break clients that stream, and metering the proxied body is the reservation design #839 rejected. The overshoot is bounded by the usage cache interval and is documented as the best-effort contract.
- Exporter cache after `Update`/`Delete`: the exporter cache lives for one scrape interval and the controller cannot import the exporter without a cycle. A quota change shows up on the next scrape.
- Keeping `hard`/`enforce` gauges when the usage query fails: that only happens when the database is unavailable, and then every other collector is missing too.
- `accounted` = `SUM(blob.size)` + system artifacts: this is exactly how `GET /statistics` computes `total_storage_consumption`, and the two numbers must agree. Docs say so now.
- `COALESCE` on the aggregate: already present.
- A separate editor component for the global quota: the shared modal gained one mode flag and one checkbox; splitting it duplicates the unit/validation logic. Can be revisited if a third mode appears.

bupd:
Triaged this rather than deep-tested it, so treat the below as review notes with one concrete measurement request attached — I have not run the suite on this branch yet.

Two things I want to flag before it merges, both about the hot path rather than the feature.

## Every install pays for this, including installs with no quota set

`Gate` → `CheckCapacity` → `controller.quota()` → `cache.FetchOrSave`. `src/lib/cache/helper.go:57-59` does **not** cache builder errors, and `dao.Get` returns `NotFoundError` when the `system_quota` singleton row is absent. So on an install that never configures a global quota, every manifest PUT and every blob POST/PATCH/PUT pays a Redis miss plus a `SELECT` on `system_quota`, forever. `fetchOrSaveGroup`'s singleflight dedups concurrent pushes within a process but doesn't remove the per-request round trip.

Caching the negative result — a nil sentinel for `quotaCacheTTL` — would make the disabled case free, which is what it should be given the middleware is mounted on four registry write routes unconditionally.

## Monolithic blob uploads aren't counted

`isBlobUpload` requires `PATCH` or `PUT`. A monolithic `POST /v2/<repo>/blobs/uploads/?digest=…` carries the entire blob body but gets `extra = 0`, so it's checked against current usage and never against usage + incoming size. `crane` and some CI clients use that path for small blobs.

## What I checked and liked

The schema side is clean. `system_quota` goes into `harbor_next.sql` (not a numbered migration), it's `CREATE TABLE IF NOT EXISTS` with a `CHECK (id = 1)` singleton guard, and it has no foreign keys — so it needs no stub in `authoritative_db_test.go`, and correctly doesn't add one.

The accounting is also cleaner than I expected: there's deliberately no reservation and no rollback, `Gate` is a pure read of cached state, and it sits *before* `quota.PutManifestMiddleware()` / `quota.PostInitiateBlobUploadMiddleware()` in `src/server/registry/route.go`, so a global denial short-circuits before any per-project reserve happens. Nothing to double-count, nothing to roll back.

That does mean overshoot is bounded by `accountedCacheTTL = 5m` rather than by any reservation — concurrent pushes inside one cache window can collectively exceed `hard`. The PR treats this as best-effort, which is a defensible design for an instance-wide cap, but `docs/global-quota.md` should say so in those words rather than leaving operators to infer it.

## Not tested

I have not run `./controller/systemquota/...`, `./server/middleware/systemquota/...`, the `-tags db` DAO suite, or measured the per-push cost described above. Happy to do the measurement if useful — the interesting number is p99 latency on a blob PATCH with the feature unconfigured, against `main` as the baseline.

Not a 2.15.9 candidate either way: new table, new public API, new RBAC resource.

PR #848, opened 2026-09-08
feat: Report Registry Volume Usage via registryctl

Vad1mo:
## Summary

Theme B of the decision record #839: Harbor learns the real size and usage of its registry storage volume. Stacked on #845 (theme A) because it installs the measurer that turns the global quota's usage from `accounted` into `measured`.

- registryctl gains `GET /api/registry/storage`: `statfs` of the registry's `rootdirectory` for the `filesystem` driver, `supported: false` for object stores. registryctl is the only component that shares the registry's mount, so this is the one place a truthful measurement can run.
- Core **pulls** it through the existing registryctl client (`Storage()`), cached for one minute via the shared cache, and degrades to its local disk with `measured: false` when registryctl is unreachable or the driver is unsupported. Core now initializes the registryctl client at startup.
- `GET /api/v2.0/systeminfo/volumes` returns `used`, `driver`, `measured`, `measured_at`; the swagger no longer claims it "only reflects local disk".
- Global quota: `systemquota.SetMeasurer(systeminfo.NewMeasurer())`, so `used_source` becomes `measured` and enforcement uses the physical value where available.
- Exporter gauges `harbor_storage_volume_bytes{kind,driver}` and `harbor_storage_volume_measured` (in-core mode).
- Docs: `docs/storage-usage-reporting.md`.

No Helm or Compose change: registryctl already mounts the volume and core already has `REGISTRY_CONTROLLER_URL` and the secret registryctl accepts.

## Related Issues

Implements #839 (theme B). Depends on #845.

## Type of Change

- [x] New feature (`feat:`)

## Release Notes

`GET /api/v2.0/systeminfo/volumes` now reports the **real registry storage volume** on Kubernetes and Compose installs using filesystem storage: total, free and used bytes measured by registryctl on the registry's own mount, with `measured: true`. Previously the numbers described the core container's disk. Where a global storage quota is set, its usage switches from Harbor's blob accounting to the measured value automatically. Object storage backends are not measured yet and keep reporting `measured: false`. New Prometheus gauges `harbor_storage_volume_bytes` and `harbor_storage_volume_measured` are exposed by the in-core exporter. Nothing needs to be configured. See `docs/storage-usage-reporting.md`.

## Testing

- [x] Unit tests added/updated (registryctl storage handler, registryctl client, core registryctl driver incl. fallback and caching, measurer adapter, exporter collector)

Local runs: `go test` for all touched packages, `golangci-lint` (0 issues), spectral API lint (0 errors).

## Checklist

### DIFF
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -8314,6 +8314,7 @@ definitions:
       measured_at:
         type: string
         format: date-time
+        x-nullable: true
         description: When the volume was measured. Only set when measured is true.
   GeneralInfo:
     type: object
--- a/api/v2.0/swagger.yaml
+++ b/api/v2.0/swagger.yaml
@@ -4264,10 +4264,13 @@ paths:
           $ref: '#/responses/500'
   /systeminfo/volumes:
     get:
-      summary: Get system volume info (total/free size).
+      summary: Get system volume info (total/free/used size).
       operationId: getVolumes
       description: |
-        This endpoint is for retrieving system volume info that only provides for admin user.  Note that the response only reflects the storage status of local disk.
+        Get the total, free and used size of the registry storage volume. Only system administrators can call this API.
+        The numbers are measured by registryctl on the registry's own volume for the filesystem storage driver; for
+        other drivers, or when registryctl cannot be reached, they fall back to the local disk of the core service and
+        `measured` is false.
       tags:
         - systeminfo
       parameters:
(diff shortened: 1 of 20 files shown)


Vad1mo:
Review pass over the cubic threads. Addressed in the latest commit.

**Fixed**
- `Storage()` takes a context and the driver bounds the round trip to five seconds; a registryctl that accepts the connection but never answers now falls back instead of blocking the first quota-gated push (test `TestHungRegistryctlFallsBackWithinTimeout`).
- A missing `rootdirectory` is answered with an error, not a zero-sized "measurement", so core never trusts `used=0`.
- The fallback capacity derives `used` from `total-free`.
- An unsupported driver (object store) reports its real driver name with zero sizes and `measured: false`; core's local disk is no longer attributed to S3.
- `measured_at` is `x-nullable` and omitted unless measured.
- The exporter publishes `harbor_storage_volume_bytes` only for measured values; the standalone exporter therefore never labels its own disk as the registry volume, it just reports `harbor_storage_volume_measured 0`.
- `ResetForTest` removed from the runtime package (it was unused).
- Tests: client success, non-2xx, malformed body, transport failure and canceled context; driver hung/unsupported/fallback-used; measurer nil capacity and int64 overflow; collector unmeasured and error paths with cache hygiene.

PR #869, opened 2026-09-11
perf(blob): Skip redundant blob Touch when blob is already StatusNone (backport #397)

bupd:
Two commits. The first is the backport of #397 to `release-2.15`, byte-identical to the commit on `main` (`2004d1981d`); original author Bojan Zelic. The second is a review follow-up that bounds the skip — see "Review follow-up" below.

**Reviewers: this is the one backport in the 2.15.9 set that I want a second opinion on.** It is labelled `perf`, but it changes GC interaction, and a mistake there means collecting a live blob. Please check the reasoning below rather than just the tests.

> **Update:** cubic raised a P1 on exactly that, and it held up in part. The second commit fixes it; [full analysis and reproduction here](https://github.com/container-registry/harbor-next/pull/869#issuecomment-5631456473). The "Risk" section below is superseded by "Review follow-up".

## What

`probeBlob` skips the `Touch` call when the blob row is already `StatusNone` and its `update_time` is newer than half the GC time window. If the window is unset or non-positive, the old unconditional `Touch` is kept.

## Why

Every `PUT manifest` currently calls `Touch` on the manifest blob even when the row is already `StatusNone`. `Touch` is a single-row `UPDATE blob SET version = version + 1, update_time = now(), status = 'none' ...` that takes a row lock, so pushes that share a manifest blob serialise behind each other. On a busy registry this showed up as a manifest-PUT P99 in the tens of minutes.

`config.GetGCTimeWindow` and the `config`/`time` imports already exist on `release-2.15` in `src/server/middleware/blob/util.go`, so the change applies unmodified.

## Risk

This is the judgement call. The safety argument is that GC only considers blobs older than the **full** `GC_TIME_WINDOW_HOURS` window, while the skip only applies to blobs touched within **half** that window — so there is a full half-window of margin between "we declined to refresh this row" and "GC would consider it". The `window <= 0` branch preserves today's behaviour when the window is unconfigured.

If that margin argument does not hold in some deployment (for instance a very short configured window combined with a long-running push), the failure mode is severe: a live blob becoming GC-eligible. I read the logic and believe it holds, but it deserves someone who knows the GC path.

The conservative alternative is to defer this one to 2.16 and ship 2.15.9 with only #867 and #868. That costs nothing except leaving the push-latency pathology in place for 2.15 users.

## How verified

- `jj duplicate` onto `release-2.15` applied with no conflict; resulting diff byte-identical to `2004d1981d`.
- The upstream change ships 205 lines of new tests across `head_blob_test.go`, `put_blob_upload_test.go`, `util_test.go` and `util_db_test.go`, including the skip/no-skip boundary; they come with the cherry-pick and CI on this PR runs them.

## Review follow-up (commit 2)

`Touch` never makes a blob ineligible for GC — its only effect on the candidate query

```sql
WHERE pb.id IS NULL AND b.update_time <= now() - interval 'N hours'
```

is moving `update_time` forward. So the pre-#397 behaviour was a timer of length `window` started at `probeBlob`, not protection-until-association, and #397 replaced it with a timer whose floor is `window/2`. Halved margin, not a new failure mode.

The follow-up bounds the skip by a short absolute age (`touchSkipMaxAge = 5m`) instead of `window/2`, so the margin is `window - 5m` and no longer shrinks with the window. The contention win survives because concurrent re-pushes of one digest arrive seconds apart.

Verified three ways:

1. **The GC predicate, against a real Postgres.** Seeded unassociated `StatusNone` blobs at 5-minute intervals and ran the exact SQL from `dao.go:413`. Youngest eligible blob: 60 min at a 1h window, 120 min at 2h, 240 min at 4h — the boundary is exactly `update_time <= now() - window`. A 30-day-old blob *with* a `project_blob` row returns 0 rows, confirming association is total immunity.
2. **A property test that fails on the unfixed code.** `TestShouldTouchNoneLivenessMargin` walks every age across several windows and asserts the worst margin an accepted skip can leave. Against unmodified #397 it reports `31m0s` vs `55m0s` at 1h and `1h1m0s` vs `1h55m0s` at 2h; it passes with the fix.
3. `go vet` and the full `./server/middleware/blob/` package suite pass.

Two details of the P1 that did not hold, recorded so they are not re-raised: the association is **not** post-response (`middleware.AfterResponse` buffers and can `Reset()`, so a failure surfaces as a retryable push error, not a 201 over a missing blob); and "keep the blob protected until association" describes a mechanism upstream does not have — the GC time window *is* Harbor's entire answer to that gap.

**If you would rather ship 2.15.9 without any of this**, dropping the PR costs only leaving the push-latency pathology in place; nothing else in the release depends on it.

## Note for `main`

`main` carries the unmodified #397 logic and has the same halved margin. If this approach is accepted here I will open the equivalent one-commit PR against `main` so the two lines do not diverge.

### DIFF
--- a/src/server/middleware/blob/head_blob.go
+++ b/src/server/middleware/blob/head_blob.go
@@ -55,7 +55,15 @@ func handleHead(req *http.Request) error {
 	}
 
 	switch bb.Status {
-	case blob_models.StatusNone, blob_models.StatusDelete:
+	case blob_models.StatusNone:
+		if shouldTouchNone(bb) {
+			if err := blob.Ctl.Touch(req.Context(), bb); err != nil {
+				log.Errorf("failed to update blob: %s status to StatusNone, error:%v", blobInfo.Digest, err)
+				return errors.Wrapf(err, "the request id is: %s", req.Header.Get(requestid.HeaderXRequestID))
+			}
+		}
+	case blob_models.StatusDelete:
+		// GC marked this blob as a candidate: rescue it back to StatusNone.
 		if err := blob.Ctl.Touch(req.Context(), bb); err != nil {
 			log.Errorf("failed to update blob: %s status to StatusNone, error:%v", blobInfo.Digest, err)
 			return errors.Wrapf(err, "the request id is: %s", req.Header.Get(requestid.HeaderXRequestID))
--- a/src/server/middleware/blob/head_blob_test.go
+++ b/src/server/middleware/blob/head_blob_test.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	beego_orm "github.com/beego/beego/v2/client/orm"
 	"github.com/stretchr/testify/suite"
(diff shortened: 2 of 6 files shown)


bupd:
@coderabbitai review

Auto review is off for non-default base branches, and this is the one backport in the 2.15.9 set that warrants a look. Please focus on the GC interaction in `src/server/middleware/blob/util.go`: the change skips the `Touch` row update when a blob is already `StatusNone` and its `update_time` falls inside half the configured GC time window.

The safety claim is that GC only considers blobs older than the **full** `GC_TIME_WINDOW_HOURS` window, so declining to refresh a row inside half that window leaves a full half-window of margin before the blob could become collectible. The `window <= 0` path keeps the old unconditional `Touch`.

What I want checked: whether that margin can be defeated — a short configured window, a long-running push, clock skew between the API and the GC job, or any GC path that uses a different threshold than the one assumed here. The failure mode if the reasoning is wrong is collecting a live blob, which is why I am asking rather than relying on the tests.

bupd:
Fixed in 76ed721 — the P1's mechanism is real, though two details of the write-up are not. I traced the GC path rather than working from the comment above the function.

### What GC actually does

`pkg/blob/dao.GetBlobsNotRefedByProjectBlob` (`src/pkg/blob/dao/dao.go:413`):

```sql
SELECT b.id, b.digest, b.content_type, b.status, b.version, b.size
FROM blob AS b LEFT JOIN project_blob pb ON b.id = pb.blob_id
WHERE pb.id IS NULL AND b.update_time <= now() - interval '%d hours';
```

The important consequence: **`Touch` never makes a blob ineligible.** Its only effect on this predicate is moving `update_time` forward. So the pre-#397 behaviour was not "protection until association" — it was a timer of length `window` started at `probeBlob`, and a push outliving it was already collectable. #397 did not open a hole; it shortened the fuse.

I reproduced the predicate against a real Postgres, seeding unassociated `StatusNone` blobs at five-minute intervals and running that exact SQL:

| window | youngest eligible blob |
|---|---|
| 1h | 60 min |
| 2h | 120 min |
| 4h | 240 min |

and a 30-day-old blob **with** a `project_blob` row returns 0 rows — association is total immunity, which is why the time window only has to bridge the gap until association.

### The actual regression

Let `W` = window, `s` = the row's age at `probeBlob`.

- **Before #397**: `touch()` unconditional → eligible at `now + W`. Margin **`W`**.
- **With #397**: skip requires `s ≤ W/2` → eligible at `update_time + W` = `now + (W − s)`. Margin **`[W/2, W)`, floor `W/2`**.

At the default `W=2h` the floor is 1h; at the smallest window the hours-granular `GC_TIME_WINDOW_HOURS` allows, 30 min. Worth noting the exposure is widest not on `put_manifest` — nginx caps a single request at `proxy_read_timeout 900` — but on `head_blob.go`, which got the same skip and where the protection has to span an entire `docker push`, not one request.

### The fix

Bound the skip by a short absolute age instead of a fraction of the window:

```go
const touchSkipMaxAge = 5 * time.Minute

func shouldTouchNone(bb *models.Blob) bool {
	window := time.Duration(config.GetGCTimeWindow()) * time.Hour
	if window <= 2*touchSkipMaxAge {
		return true
	}
	return time.Since(bb.UpdateTime) > touchSkipMaxAge
}
```

Margin becomes `W − 5min` instead of `W/2`, and stops shrinking as `W` shrinks. The contention win is kept: concurrent re-pushes of one digest arrive seconds apart, so a five-minute skip absorbs the burst that produced the row-lock hot spot. The skips this gives up — a push arriving 40 minutes after the last one — were not contending with anything.

`TestShouldTouchNoneLivenessMargin` pins the property rather than the arithmetic: it walks every age across several windows and asserts the worst margin any accepted skip can leave. Against the unmodified #397 logic it fails exactly as predicted —

```
window=1h: "31m0s" is not greater than or equal to "55m0s"
window=2h: "1h1m0s" is not greater than or equal to "1h55m0s"
window=6h: "3h1m0s" is not greater than or equal to "5h55m0s"
window=24h: "12h1m0s" is not greater than or equal to "23h55m0s"
```

— and passes with the fix.

### Two corrections

**"before the post-response association commits"** — the association is not post-response. `middleware.AfterResponse` (`src/server/middleware/middleware.go:75-90`) runs the hook against a buffered `lib.ResponseBuffer` and, on hook error, does `res.Reset()` + `SendError`. `AssociateWithProjectByDigest` / `AssociateWithArtifact` therefore complete before the client sees the status line, and a failure surfaces as a retryable push error rather than a 201 over a missing blob.

**"keep the blob protected until association"** — no such mechanism exists upstream. Nothing in GC consults in-flight upload sessions; the `GC_TIME_WINDOW_HOURS` window *is* Harbor's entire answer to the gap between blob upload and manifest association (see the "reserve window" comment at `src/pkg/blob/dao/dao.go:82`). Asking the skip to provide it would mean building something the unpatched code does not have either.

Also worth recording for anyone reading this later: the `StatusDelete` / `StatusDeleteFailed` rescue path still touches unconditionally — the skip is confined to the `StatusNone` branch — so the mark→sweep CAS (`version = ?` in `UpdateBlobStatus`) is untouched by this change and remains the second line of defence.

bupd:
Closing: 2.15.9 carries P0 fixes only, and this is a performance change, not a P0.

Two further reasons it is the right one to drop, both surfaced during review:

- It would have been the only commit on `release-2.15` deliberately diverging from upstream #397, because shipping upstream's logic unmodified halves the GC liveness margin (`W` → `W/2`, so 30 min at the smallest window `GC_TIME_WINDOW_HOURS` allows). Carrying a 2.15-only variant of a GC-adjacent change through a patch release is exactly the sort of thing a patch release should not do.
- The cost of dropping it is only that the push-latency pathology stays in place for 2.15 users. Nothing else in the release depends on it, and `release-2.15` is unaffected — this never merged.

**The finding itself is not closed with the PR.** `main` still carries the unmodified #397 logic and the same halved margin. The analysis, the Postgres reproduction of the GC predicate, and the property test that fails against the unfixed code are all in [this comment](https://github.com/container-registry/harbor-next/pull/869#issuecomment-5631456473) and in commit `76ed721` on this branch, so a `main`-targeted PR can be lifted from here whenever it is wanted. Branch `backport-397-blob-touch-skip-2.15` is left in place for that.

Superseded, not rejected.

PR #870, opened 2026-09-11
ci: Take commercial release notes from the patch changelogs

bupd:
## Summary

Backport of #843 to `release-2.15`, with the feature-name inventory dropped. Companion to
#896, which makes the same change on `main`.

`## Commercial Features` stops listing the tip commit subject of every branch in
`taskfile/commercial-patches` under "This release includes the following commercial
enhancements", and starts rendering what each branch's `changelogs/<branch>.md` says
changed since the previous release marker.

#793 comes with it: #843 renders its `## Summary` from the same output path, and this
branch has no `extract-pr-summary.mjs` at all, so the engine would call a script that does
not exist and abort under `set -e`.

## How the window works

Entries come from the **unreleased block** of each branch's changelog — the lines above
the newest `--- release vX.Y.Z (target <sha12>) ---` marker — because notes render before
the release cut stamps one. Re-rendering an old tag reads that tag's own section instead.
A tag older than every marker renders nothing.

## Differences from #843 as it stands on main

- **No feature-name inventory.** #843 collapses unchanged features into
  `_No changes this release: A, B, C._`; that line is gone. A release that touched no
  commercial feature renders no section at all.
- **A branch with no `# Title` falls back to its own branch name** and warns, rather than
  to `git log -1 --format=%s`. Under merge-sync every tip subject is a bot merge, so that
  fallback prints `sync: merge main … into 0001-commercial-feature-gate` as a feature
  name. All six branches on this line warn today; nothing leaks into the output.
- **Feature blocks are separated by a blank line** — #871's fix, folded in so the two
  lines render identically.

## Backport entries can now carry a release note

`extract-pr-summary.mjs` finds a PR through the canonical `in [#N](.../pull/N)` reference,
synthesized from release-please's own `(#N)` suffix. A backport subject already ends in a
parenthesis — `fix(scan): … (backport #863)` — so release-please appends nothing and no
reference is synthesized. The fallback reads the pull request the subject names.

Without it v2.15.9 renders an empty `## Summary`, because every entry in it is a backport
or an upstream pick. With it, #863's own release note reaches a 2.15.x release for the
first time:

```
## Summary

A re-scan that yields an empty raw report no longer clears the artifact's vulnerability
associations while leaving its severity counters at the previous scan's values. The last
successful result is kept and the scan is marked failed.
```

#716 still contributes nothing, because its description has no `## Release Notes` section.
That is a content gap, not a tooling one, and descriptions stay editable after merge.

## Verification

[...]

PR #871, opened 2026-09-11
ci: Separate commercial feature blocks in release notes

bupd:
One-line spacing fix in the commercial-features renderer added by #843.

## What

`update-release-notes.sh` writes each feature block as:

```sh
echo "### ${feature_title}"
echo
printf '%s' "${feature_entries}"
echo
```

`feature_entries` comes from a command substitution, so its trailing newline is already gone. The bare `echo` restores exactly that one newline and nothing more, which means the next `### Feature` heading — and the trailing `_No changes this release: ..._` line — start on the line immediately after the last bullet.

Rendered against a branch set where two features have entries and one does not:

```
### Branding customization

- Custom login logos now survive an upgrade (#820)
  Previously the logo reset to the Harbor default.
### Federated Robot Accounts

- Renamed identity providers (#818)
_No changes this release: Commercial Feature Gate._
```

GitHub treats that last italic line as a lazy continuation of the preceding list item, so it renders inside the bullet instead of as its own paragraph.

`printf '%s\n\n'` emits the newline the substitution stripped plus the blank line that separates the block from whatever follows, and the trailing `echo` goes away.

## Why now

Not visible on any released page yet: the renderer only reaches this branch when a patch branch has changelog entries, and none of the `changelogs/<branch>.md` files are seeded. It would have surfaced on the first release after seeding.

Found while backporting #843 to `release-2.15` (#870), which carries the same fix so the two branches do not diverge.

## Risk

None beyond the notes themselves. The engine runs at release time only, and notes can be regenerated with the `Recreate Release Notes` workflow.

## How verified

Reproduced the unfixed output above against the current script, then confirmed the fix restores a blank line before each subsequent heading and before the trailing italic line. `bash -n` clean.

### DIFF
--- a/.github/scripts/update-release-notes.sh
+++ b/.github/scripts/update-release-notes.sh
@@ -247,8 +247,10 @@ if [[ "${chart_mode}" == false && -f "${series}" ]]; then
       {
         echo "### ${feature_title}"
         echo
-        printf '%s' "${feature_entries}"
-        echo
+        # printf keeps the entries verbatim; the two newlines restore the one
+        # command substitution stripped plus the blank line that separates
+        # this block from whatever follows it.
+        printf '%s\n\n' "${feature_entries}"
       } >> "${patch_notes}"
     else
       unchanged_features+=("${feature_title}")


bupd:
Superseded by #896 — recommending closure, not closing.

This PR fixes the spacing of the `## Commercial Features` block added by #843. #896 deletes that block: release notes are taken from the changelog now, and the commercial feature list goes with the machinery that produced it, because the changelog already records what shipped.

The specific bug here is real and I re-derived it independently — `printf '%s' "${feature_entries}"` followed by a bare `echo` restores exactly the one newline the command substitution stripped, so the next `### Feature` heading and the trailing `_No changes this release: …_` line render as a lazy continuation inside the last bullet. The fix in this PR is correct. There is just no longer a block for it to apply to: after #896 the emit site at `update-release-notes.sh:265-280` does not exist.

Worth recording that the block never rendered a real entry in production either way. `git ls-tree` on the live `0001`, `0002`, `0005` and `0007` patch branches shows `changelogs/` empty on all four, so #843's reader has only ever run its fall-back-to-tip-subject path and every feature always landed in the `_No changes_` line the spacing bug affected.

@bupd — your call on closing.

bupd:
Retracting my previous comment — I had the design backwards and this PR is **not** superseded.

I read "take the release notes from the changelog" as "delete the commercial section." It means the opposite: take the commercial entries from `changelogs/<branch>.md` on the patch branches declared in `taskfile/commercial-patches`, instead of from the branch tip commit subjects. The block stays. What goes is the static inventory of feature names, because the changelogs already carry the actual changes.

So the spacing bug this PR fixes is live and still matters — the `printf '%s'` + bare `echo` leaves a feature's entries with no blank line after them, and the next `### Feature` heading renders as a lazy continuation inside the last bullet. That is exactly the block that survives.

Please disregard the closure recommendation. I am folding this fix into the reworked #870/#896 so the two lines render identically; if you would rather keep it as its own PR, say so and I will drop it from mine and rebase this one instead.

Sorry for the noise.

bupd:
Superseded by #896, which carries this spacing fix plus the changelog-sourced commercial section. The two would conflict on merge.

PR #874, opened 2026-09-11
fix(api): Reject query operands the column cannot take

bupd:
>
> 2. #874 — query operand validation (base `fix/islocalharbor-core-url-normalize`) **→ this PR**
>

## What

A query filter operand that does not match its column type reached Postgres verbatim, and the driver error came back to the client as a 500.

```
GET /projects?q=creation_time=[a~b]
→ 500 UNKNOWN / internal server error
core: ERROR: invalid input syntax for type timestamp: "a" (SQLSTATE 22007)
```

Same for `/audit-logs?q=op_time=[x~y]` and `/artifacts?q=push_time=[2020~abc]`. gh#23670 ("prevent 500 on invalid query input") only hardened the type assertions in the robot handler, so this class was never covered.

## Wider than the report

Reproduced at the DAO layer against a real Postgres. Ranges are not the only broken form — exact matches and or-lists fail the same way, and integer columns fail alongside timestamp ones:

```
q="creation_time=[a~b]"   -> ERROR: invalid input syntax for type timestamp: "a" (SQLSTATE 22007)
q="creation_time=[2020~abc]" -> ERROR: invalid input syntax for type timestamp: "2020" (SQLSTATE 22007)
q="creation_time=abc"     -> ERROR: invalid input syntax for type timestamp: "abc" (SQLSTATE 22007)
q="creation_time={a b}"   -> ERROR: invalid input syntax for type timestamp: "a" (SQLSTATE 22007)
q="project_id=[a~b]"      -> ERROR: invalid input syntax for type integer: "a" (SQLSTATE 22P02)
q="project_id=abc"        -> ERROR: invalid input syntax for type integer: "abc" (SQLSTATE 22P02)
q="creation_time=~abc"    -> no error   (ILIKE casts the column to text)
q="name=[a~b]"            -> no error   (text column)
```

## Fix

`parseModel` already walks the model's struct fields to decide what is filterable and sortable; it now records each field's Go type too. `setFilters` checks the operand against that type before applying the filter and returns a `BadRequestCode` error instead, which the handler renders as 400.

The check is deliberately permissive so nothing that works today stops working:

- text columns take any literal
- integer and float columns take numeric values and numeric strings (`"12"` still filters, Postgres casts it)
- timestamp columns take a `time.Time` or a string in a layout Postgres accepts, so a date-only bound like `creation_time=[2020-01-01~2021-01-01]` keeps working
- fuzzy matches are not checked at all — they render as ILIKE, which casts the column to text
- and-lists are not checked; the DAO builds those conditions itself
- list operands (the values behind `__in`) are checked element by element

This covers every endpoint that filters through `orm.QuerySetter`, not just the three in the report.

## Tests

`TestValidateFilterValue` (pure) covers the matrix above plus pointer fields, bool columns, slices, and nil. `TestListInvalidFilterValue` / `TestListValidFilterValue` (db-tagged, `pkg/project/dao`) run the same queries through a real Postgres and assert 400 vs success.

```
$ go test ./lib/orm/ ./lib/q/
ok  	github.com/goharbor/harbor/src/lib/orm
ok  	github.com/goharbor/harbor/src/lib/q

$ go test -tags db ./pkg/project/dao/ -run 'TestDaoTestSuite/TestList(Invalid|Valid)FilterValue' -v
--- PASS: TestDaoTestSuite/TestListInvalidFilterValue (6 subtests, all bad_request=true)
--- PASS: TestDaoTestSuite/TestListValidFilterValue  (6 subtests, all nil error)
```

The full db-tagged lane (103 packages) was run locally against Postgres 16. Three packages fail — `jobservice/period`, `pkg/blob`, `pkg/blob/dao` — and they fail identically on `main` with this change reverted, on a freshly created database. Not caused by this PR.

## CI note


bupd:
## Post-fix sweep: what this PR does and does not cover

I swept the whole v2.0 API against a local dev stack running this branch (`fix874`, ORM operand validation applied) to find 500s that survive the fix. Targets came from `api/v2.0/swagger.yaml`: all 42 GET operations taking `q`, `sort`, `page` or `page_size`. Roughly 26k sequential requests, admin and anonymous, covering malformed operands per column type, malformed `q` syntax, bad `sort` keys, bad paging and each endpoint's own typed params.

**The fix holds for everything it aims at.** Every case in the issue table now returns 400 or 422 instead of 500, including `/projects?q=creation_time=[a~b]`, `/audit-logs?q=op_time=[x~y]` and `/system/purgeaudit?q=update_time=abc`. The sort path is clean on every endpoint (`setSorts` drops unsortable keys), and paging is clean (go-swagger rejects bad `page` / `page_size` with 422 before the handler runs). Core restart count stayed 0 throughout.

**Five residual classes remain.** None of them are regressions from this PR and none are in scope for it, but one is the same failure mode reaching columns `validateFilterValue` cannot see:

1. **json / jsonb columns.** `validateFilterValue` keys off the Go struct field type. `Annotations` and `ExtraAttrs` are declared `string` in Go while the column is `json` / `jsonb`, so a string operand passes and Postgres still refuses the comparison. `/system/gc?q=extra_attrs=abc` → 500 `operator does not exist: json = unknown (SQLSTATE 42883)`; `/projects/library/artifacts?q=annotations=abc` → 500 `invalid input syntax for type json (SQLSTATE 22P02)`. The artifact one is reachable unauthenticated on a public project. This is the closest thing to a gap in this PR: the field type is not enough to decide, the orm tag's `type(...)` would be needed for the artifact case and nothing in the struct helps for `execution.extra_attrs`.
2. **`/security/vul`.** `pkg/securityhub/dao` builds its own SQL and never calls `orm.QuerySetter`, so nothing here applies. `q=project_id=abc` and `q=cvss_score_v3=[a~b]` both 500.
3. **`/system/purgeaudit` nil deref.** `purge.go:158-162` writes `query.Keywords[...]` before checking the `BuildQuery` error, so any unparseable `q` panics before the ORM is reached. `q==`, `q=id=[a~`, `q=name=` all return a bare 500 with an empty body.
4. **`latest_in_repository=true`.** `artifact/dao.ListWithLatest` binds the parsed `media_type` / `artifact_type` operand straight into raw SQL, so `[a~b]`, `{a b}`, `(a b)` and `~a` panic inside beego.
5. **Retention execution type assertion.** `controller/retention.convertExecution` does `ExtraAttrs["dry_run"].(bool)` unchecked and `GetRetentionExec` has no vendor filter, so `/retentions/1/executions/1/tasks` returns 500 on a stock install. Not a query-filter bug, found while sweeping paging.

All five code paths exist unchanged on goharbor/harbor `main` at `3488b6455a7a8cde5cbdf92b243b54567b9e61a9`. Full repro matrix and code references are in #912, and I posted the upstream-relevant parts to goharbor/harbor#23916: https://github.com/goharbor/harbor/issues/23916#issuecomment-5661672281

bupd:
Re-checked against gap 1 of the 2.15.8 verification pass (timestamp RANGE operands still 500). **This PR already covers it** — no code change needed, so it stays as reviewed.

`validateFilterValue` has an explicit `*q.Range` case that validates both bounds through `validateOperand`, and `timestampAssignable` rejects anything that no accepted layout parses. All three reported probes land on that path:

| probe | operand | result |
| --- | --- | --- |
| `GET /projects?q=creation_time=[a~b]` | `Range{Min:"a", Max:"b"}` | 400 |
| `GET /audit-logs?q=op_time=[x~y]` | `Range{Min:"x", Max:"y"}` | 400 |
| `GET /artifacts?q=push_time=[2020~abc]` | `Range{Min:"2020", Max:"abc"}` | 400 |

`2020` is rejected too — it matches none of the layouts in `timestampLayouts`, so the bad lower bound is caught even though the upper one is the obviously broken half.

All three endpoints filter through `orm.QuerySetter`, which is where the check sits: `pkg/project/dao/dao.go:162`, `pkg/auditext/dao/dao.go:77` (and `pkg/audit/dao/dao.go:151`), `pkg/artifact/dao/dao.go:335`. `TestValidateFilterValue` already pins the range cases, and `TestListInvalidFilterValue` runs them through a real Postgres.

Two administrative changes, no diff impact:

- The base is back to `main` (GitHub retargeted it when #873 merged), so full CI runs now. The "only branch-agnostic checks run while stacked" note in the description no longer applies.
- The rest of the stack is gone. #875 is closed and replaced by #923, #876 by #924, both independent off `main` so each gets its own preview images for isolated verification.

bupd:
Pushed a fixup: `pkg/project/dao/dao_test.go` declared `TestListValidFilterValue` twice, so the db-tagged lane failed to build and Unit Tests (DB 1/2) was red:

```
pkg/project/dao/dao_test.go:475:28: method DaoTestSuite.TestListValidFilterValue
  already declared at pkg/project/dao/dao_test.go:438:28
```

The second copy arrived with "Cover the permissive path at the DAO layer" and is a superset of the first — it adds the fuzzy `creation_time=~2020` and the integer range `project_id=[1~2]` cases. Dropped the earlier copy and kept the superset. `go vet -tags db ./pkg/project/dao/` is clean now.

The previous run was also why the image matrix was skipped, so no preview images matching this head existed. This push should rebuild them as `pr-874`.

bupd:
Superseded by #925 — same diff, plus the duplicate-test build fix already pushed here.

Closing only because this PR cannot produce preview images. It is position 2 of 4 in GitHub stack #907, and `pr-ci.yml` builds the image matrix only for the top of a stack:

```yaml
if: >-
  (github.event.pull_request.stack == null ||
   github.event.pull_request.stack.position == github.event.pull_request.stack.size)
```

The top of stack #907 is #876, which is closed, so nothing in this stack will ever build images again. Neither the REST nor the GraphQL API can retarget or detach a stacked PR ("Cannot change the base branch because the pull request is part of a stack"), and GitHub exposes no detach mutation. Since each of the three release-blocking fixes needs its own preview images for isolated verification, a fresh PR was the only route.

The code review here still stands — #925 is the identical diff on a branch outside the stack.

PR #875, opened 2026-09-11
fix(registry): Answer 400 when the endpoint health check cannot reach the URL

bupd:
>
> 3. #875 — registry health check 400 (base `fix/query-filter-value-validation`) **→ this PR**
>

## What

```
POST /registries {"url": "ftp://example.invalid", ...}  → 500
POST /registries {"url": "not-a-url", ...}              → 500
```

Core log:

```
Get "ftp://example.invalid/api/version": unsupported protocol scheme "ftp"
Get "http://not-a-url/api/version": dial tcp: lookup not-a-url ... no such host
```

## Which of the two it actually is

The brief asked whether `lib.ValidateURL` is simply not called on this path, or is called and the later health check is what 500s. It is the second.

`controller.validate` does call `lib.ValidateURL(registry.URL)` (src/controller/registry/controller.go:95) and both inputs pass it:

- `ftp` is in `ValidateURL`'s default scheme set (`http, https, s3, sftp, ftp`)
- a scheme-less string is normalised to `http://not-a-url`, which is a structurally valid URL

So the request reaches `IsHealthy`, and the transport error from `adapter.HealthCheck()` is returned untouched. `validate` already answers 400 when the adapter reports *unhealthy* — the inconsistency is that failing to reach the endpoint at all was a 500.

## Fix

Wrap the `HealthCheck` error as a bad request, keeping the underlying reason in the message. The same path backs `POST /registries/ping` and `PUT /registries/{id}`, which returned 500 for the same input.

## What is deliberately not changed

Narrowing `ValidateURL` to `http, https` on this path would be the other obvious fix, and it is wrong here: storage-backed registry types use `s3://` and `sftp://` URLs and go through the same `validate`. `TestUpdateRegistryStorageSchemeURLAccepted` guards that, and the commercial SFTP replication patch creates registries with `sftp://` endpoints. The scheme allowlist stays as it is.

## Tests

```
$ go test ./controller/registry/ ./server/v2.0/handler/ -run 'TestRegistryTestSuite' -v
--- PASS: TestRegistryTestSuite/TestValidateHealthCheckTransportError/unsupported_scheme
--- PASS: TestRegistryTestSuite/TestValidateHealthCheckTransportError/unresolvable_host
--- PASS: TestRegistryTestSuite/TestPingRegistryUnreachableURLRejected
... plus the 10 existing subtests
ok  	github.com/goharbor/harbor/src/controller/registry
ok  	github.com/goharbor/harbor/src/server/v2.0/handler
```

## CI note


bupd:
Superseded by #923.

Two reasons this one could not be repaired in place:

1. The fix was in the wrong branch. It wrapped `HealthCheck()`, but both reported probes fail one call earlier — the Harbor adapter issues `GET /api/version` from `base.New` while `CreateAdapter` is still running. The unit test here only passed because it mocked `CreateAdapter` into succeeding, which hid the call that actually fails.
2. GitHub refuses to retarget this PR to `main` ("Cannot change the base branch because the pull request is part of a stack") via both the REST and GraphQL APIs, and exposes no mutation to detach it. The three release-blocking fixes must be independent so each gets its own preview images for isolated verification, so a fresh PR was the only way out of the stack.

#923 maps transport failures from both branches of `IsHealthy`, refuses `ftp://` up front, and carries a test that drives the real registry manager and real adapter factory instead of mocks.

PR #876, opened 2026-09-11
fix(db): Widen execution.revision to bigint

bupd:
>
> 4. #876 — execution.revision bigint (base `fix/registry-health-check-bad-request`) **→ this PR**
>

## What

`execution.revision` is `integer` while the Go model declares it `int64` (`src/pkg/task/dao/model.go:45`). gh#23718 widened three columns and left this one:

```sql
select table_name, column_name, data_type from information_schema.columns
where column_name in ('revision','status_revision','setup_timestamp');

      table_name      |   column_name   | data_type
----------------------+-----------------+-----------
 execution            | revision        | integer
 p2p_preheat_instance | setup_timestamp | bigint
 schedule             | revision        | bigint
 task                 | status_revision | bigint
```

## Correcting the finding

The integration-test report filed this as "`execution.revision` retains the 2038 overflow". That framing is wrong, and it is worth stating plainly before the fix.

The three columns gh#23718 converted all hold unix timestamps — `schedule.revision` is written from `sc.Metadata.CheckInAt` (`src/pkg/scheduler/callback.go:86`), `task.status_revision` from the job hook's revision. `execution.revision` is not a timestamp. It is the optimistic-lock counter behind

```sql
update execution set status = ?, revision = revision+1, update_time = ? where id = ? and revision = ?
```

so it starts at 0 and increments once per status transition of that one execution row. There is no 2038 deadline on it; exhausting int4 would take two billion status updates against a single execution.

It is still the odd column out against its own model, and widening it is cheap, so this PR does it — just not for the reason the report gave.

## Where the DDL goes

Per the repo rule, no new numbered migration. The statement goes in `migrations/postgresql/harbor_next.sql`, guarded on the current type so a repeat run never rewrites the table:

```
$ psql -d registry -c "select relfilenode from pg_class where relname='execution'"
20353
$ psql -d registry -f harbor_next.sql      # second application
DO
$ psql -d registry -c "select relfilenode from pg_class where relname='execution'"
20353                                       # no rewrite
```

## Verified

Against a Postgres 16 database migrated through the full numbered chain (schema 190):

```
before: execution.revision = integer
$ psql -d registry -v ON_ERROR_STOP=1 -f harbor_next.sql
after:  execution.revision = bigint
applied a second and third time: still bigint, no error
```

`TestAuthoritativeSchemaAgainstPostgreSQL` now creates an `execution` stub among the legacy dependencies and asserts the column comes out `bigint`. The stub is needed for the same reason `robot` and `project` are already there — the authoritative schema is applied into a bare schema in that test.

```
$ go test -tags db ./migration/ -run TestAuthoritativeSchema -v
--- PASS: TestAuthoritativeSchemaAgainstPostgreSQL (0.18s)
--- PASS: TestAuthoritativeSchemaPath (0.00s)
```

## CI note


bupd:
Replaced by #911. GitHub's stack feature pins this PR's base to #875's head branch and refuses base edits, so it cannot be retargeted past the skipped #875. #911 carries the identical commit rebased onto #874's head.

bupd:
Recreated as #924, off `main` and independent this time. Same fix; the commit message and description drop the Y2K38 framing, because `execution.revision` is an optimistic-locking counter (`revision = revision+1`) rather than a unix timestamp like `schedule.revision`. Details in #924.

PR #882, opened 2026-09-14
refactor: Replace gorilla/handlers with an in-tree access log handler

bupd:
First of several PRs retiring the Gorilla dependency. Inventory and full plan:
`herdr-cc/tasks/gorilla-status.md`.

## What this removes

`registryctl` imported `gorilla/handlers` for one call:

```go
h = gorilla_handlers.LoggingHandler(os.Stdout, h)
```

That decorator writes an Apache Common Log Format line per request. It is the
entire dependency — no other Gorilla handler is used anywhere in `src/`.

## Why

`gorilla/handlers` has no release since v1.5.2 (November 2023), no commit since
January 2024, and 27 open issues. To be accurate about it: **there is no CVE
here.** OSV has nothing against v1.5.2. This one is about not depending on a
dormant module for a decorator we can own, and it is the cheapest piece of the
Gorilla work, so it goes first.

The part of this effort that *is* a live security item is `gorilla/csrf`, where
`govulncheck` flags our own code and the pin in `.github/renovate.json` blocks
the fix. That lands separately and later — see the status doc, and note it has
to sequence behind the mfa stack's CSRF PR.

## The replacement

`src/registryctl/handlers/logging.go` renders the same line. It keeps
`httpsnoop` to wrap the `ResponseWriter`, which is what `gorilla/handlers` used
internally and what `otelhttp` already pulls into our build graph. Hand-rolling
that wrapper is how you silently lose `Flusher` and `Hijacker` on a handler that
later needs them, and #848 is actively adding to registryctl.

`httpsnoop` moves from `// indirect` to direct in `go.mod`. No module is added
to the binary; `gorilla/handlers` leaves it.

## Verifying the log line did not change

Ran both implementations over thirteen request shapes — host with and without a
port, IPv6, userinfo, empty `RequestURI`, HTTP/2 `CONNECT`, embedded quote,
backslash, newline, non-ASCII, control bytes, non-2xx, non-empty body — and
diffed the output with the timestamp field excluded. Twelve match byte for byte.

One differs, and it is worth stating rather than burying: DEL (U+007F) renders
as `\x7f` where gorilla wrote `\u007f`. Reaching it requires a raw control byte
in the request line, and `net/http` answers 400 to those before dispatch — I
checked this against a real server rather than assuming it:

```
DEL (0x7f) raw in URI  -> rejected by net/http, HTTP/1.1 400 Bad Request
NUL (0x00) raw in URI  -> rejected by net/http, HTTP/1.1 400 Bad Request
percent-encoded %7F    -> handler reached, RequestURI="/api/health?x=%7F"
```

Percent-encoded input arrives as printable characters and both implementations
pass it through unchanged. The divergence is documented at `escapeQuoted` and
pinned by a test, so nobody has to rediscover it.

## Checks

- `go build ./...` clean
- `go test ./registryctl/...` green
- `golangci-lint run ./registryctl/...` — 0 issues
- `task build:tidy` idempotent, no further `go.mod` / `go.sum` drift

## Follow-ups

`gorilla/mux` in jobservice, then registryctl, then the `common/utils/test`
helper; then `gorilla/csrf` to `net/http.CrossOriginProtection`. One package per
PR. The status doc carries the reasoning for each, including the two routes that
do not port to `ServeMux` cleanly.


bupd:
Stack merge order: #882 -> #883 -> #894. Squash-merge bottom-up; each merge deletes the head branch and GitHub retargets the next PR to main automatically. Do not delete a head branch by hand while its dependent PR is still open.

Issue #891, opened 2026-09-14
Zero CVE: release-2.15 baseline and remediation plan


bupd:
`release-2.15` has no CVE baseline. Nothing scans the images we publish, so we
cannot tell a real reduction from a scanner-database update. This issue records the
first one and lists every finding with the version that fixes it.

The Zero CVE pipeline that used to do this (#359, #368, #374, #375) was removed by
#548 when Renovate took over dependency bumps. #548 replaced the *remediation* half;
the *measurement* half was never replaced.

## Scope

Branch `release-2.15` @ 75fab53f4f (version 2.15.8), and the seven images published
at tag `v2.15.8`. Scanner Trivy 0.74.0, scanned 2026-09-14.

Branch pins, all behind `main`:

| Pin | release-2.15 | main |
|---|---|---|
| `GO_VERSION` | 1.26.5 | 1.26.5 |
| `ALPINE_VERSION` | 3.23 | 3.24 |
| `NGINX_VERSION` | 1.28 | 1.31 |
| `BUN_VERSION` | 1.2.13 | 1.4.0 |
| `TRIVY_VERSION` / `TRIVY_BASE_IMAGE_VERSION` | 0.71.2 | 0.74.0 |
| `HARBOR_SCANNER_TRIVY_VERSION` | v0.38.1 | v0.40.1 |
| `LPROBE_VERSION` | 0.1.9 | 0.2.0 |

## Baseline

| Image | Total | CRITICAL | HIGH | MEDIUM | LOW | UNKNOWN |
|---|---:|---:|---:|---:|---:|---:|
| harbor-core | 34 | 0 | 26 | 7 | 0 | 1 |
| harbor-jobservice | 32 | 0 | 26 | 5 | 0 | 1 |
| harbor-registryctl | 32 | 0 | 26 | 5 | 0 | 1 |
| harbor-exporter | 28 | 0 | 25 | 3 | 0 | 0 |
| harbor-portal | 121 | 2 | 25 | 40 | 54 | 0 |
| harbor-registry | 32 | 0 | 27 | 4 | 0 | 1 |
| trivy-adapter | 123 | 0 | 54 | 50 | 15 | 4 |
| **Total** | **402** | **2** | **209** | **114** | **69** | **8** |

Deduplicated: **212 distinct (CVE, package) pairs** — 2 CRITICAL, 46 HIGH,
91 MEDIUM, 69 LOW, 4 UNKNOWN. 167 have a published fix, 45 do not.

<details>
[...]

PR #892, opened 2026-09-14
chore(deps): Bump Go to 1.26.8

bupd:
Scanning the seven images we publish at `v2.15.8` turns up eight HIGH Go stdlib
advisories in every binary they ship:

| CVE | Fixed in |
|---|---|
| CVE-2026-33818 | 1.25.13, **1.26.6**, 1.27.0-rc.3 |
| CVE-2026-39821 | 1.25.13, **1.26.6**, 1.27.0-rc.3 |
| CVE-2026-46600 | **1.26.6**, 1.27.0-rc.3 |
| CVE-2026-56853 | 1.25.13, **1.26.6**, 1.27.0-rc.3 |
| CVE-2026-56858 | 1.25.13, **1.26.6**, 1.27.0-rc.3 |
| CVE-2026-56859 | 1.25.13, **1.26.6**, 1.27.0-rc.3 |
| CVE-2026-56860 | 1.25.13, **1.26.6**, 1.27.0-rc.3 |
| CVE-2026-56862 | 1.25.13, **1.26.6**, 1.27.0-rc.3 |

Every binary in every image is compiled in-tree — core, jobservice, registryctl,
exporter, registry, lprobe, scanner-trivy and the vendored Trivy CLI — so a single
`GO_VERSION` bump reaches all of them. Rows closed per image: core, jobservice,
registryctl, exporter and registry 16 each; portal 8; trivy-adapter 24. That is
**112 of the 402 findings** a full Trivy sweep reports across the seven images, from
one line.

## Which file actually controls the toolchain

`versions.env` alone would not have changed the published binaries, and the first
push of this PR got that wrong — thanks to both reviewers for catching it.

`publish-images.yml` runs `task build:binary:<component>:<platform>` **on the
runner**, after `.github/actions/setup-go-cached` resolves
`go-version-file: src/go.mod`; `actions/setup-go` then pins `GOTOOLCHAIN=local` at
that version. The production dockerfiles are `FROM scratch` and only `COPY` the
prebuilt binaries — they declare no `GO_VERSION` arg at all, so the build arg
passed at `publish-images.yml:180` never reaches a compiler.

So the `go` directive in `src/go.mod` moves too. That is the same mechanism the
earlier stdlib remediations used (#374 bumped 1.26.3 → 1.26.4, #516 bumped
1.26.4 → 1.26.5). `versions.env` moves with it to keep the dev containers on the
same toolchain. `go mod tidy` produces no further drift.

## Why 1.26.8 and not 1.27

1.26.6 is the minimum that fixes all eight; 1.26.8 is the current patch of that
line (released 2026-09-01, past Renovate's 7-day `minimumReleaseAge`).

The 1.27 line is not viable yet. Trivy's `pkg/x/json` is written against the Go 1.26
`GOEXPERIMENT=jsonv2` API, and `task build:binary:trivy:*` fails on Go 1.27.1 with:

```
# github.com/aquasecurity/trivy/pkg/x/json
pkg/x/json/json.go:109:16: undefined: json.SkipFunc
```

Worth knowing before anyone lands a 1.27 bump from a Renovate PR.

## Verification

Built the full commercial megamerge (`main` + `dev` + the eight `000N-*` patch
branches) on 1.26.8 and ran it locally under rootless podman. `go build ./...` is
clean, and every component reports healthy:

```
core=healthy database=healthy jobservice=healthy portal=healthy
redis=healthy registry=healthy registryctl=healthy trivy=healthy
```

Project creation, admin auth, the jobservice worker pool, audit logs and all seven
commercial feature gates behave as before.

Refs #891


bupd:
Both reviews are right, and the PR was a no-op as first pushed. Verified:

- `publish-images.yml:106-108` runs `.github/actions/setup-go-cached` with `go-version-file: src/go.mod`, and `actions/setup-go` pins `GOTOOLCHAIN=local` at whatever that resolves to.
- `publish-images.yml:138` then runs `task build:binary:<component>:<platform>` **on the runner**, so every shipped binary is compiled by that toolchain.
- The production dockerfiles are `FROM scratch` and only `COPY bin/linux-${TARGETARCH}/...`. None of them declares a `GO_VERSION` arg, so the build arg at line 180 never reaches a compiler.

With `go 1.26.5` in `src/go.mod`, the binaries would still have been built on 1.26.5 and kept all eight advisories.

Fixed by bumping the `go` directive to 1.26.8 as well — the same mechanism #374 (1.26.3 → 1.26.4) and #516 (1.26.4 → 1.26.5) used for earlier stdlib remediation. `versions.env` moves with it so the dev containers stay on the same toolchain. `go mod tidy` produces no further drift, and `go build ./...` is clean on 1.26.8.

I went with the `go` directive rather than adding a `toolchain` line to match that existing precedent and keep the two bumps expressed the same way.

Issue #641, opened 2026-08-18
Decouple the Trivy scanner adapter from harbor-next releases


Vad1mo:
Stop building the Trivy scanner adapter inside harbor-next, and let the Helm chart consume the images published by the [`container-registry/harbor-scanner-trivy`](https://github.com/container-registry/harbor-scanner-trivy) project instead.

The adapter is an independent component that talks to Harbor over a stable, versioned API (the Harbor scanner adapter spec). It does not need to be compiled, versioned, or released in lockstep with Harbor.

## Current situation

harbor-next builds and publishes `trivy-adapter` as if it were one of its own components:

- `versions.env` pins `HARBOR_SCANNER_TRIVY_VERSION`, `TRIVY_VERSION` and `TRIVY_BASE_IMAGE_VERSION`
- the build clones the scanner repo and the Trivy repo and cross-compiles both
- a dedicated dockerfile assembles the image
- `trivy-adapter` rides along in the PR-preview and publish image matrices
- the resulting image is tagged with the *Harbor* version

Meanwhile `harbor-scanner-trivy` already has its own release process and publishes an equivalent multi-arch image on every tag and on `main`.

## Why this should change

**Scanner updates ship on their own schedule.** Vulnerability scanning is the part of the stack that moves fastest: Trivy releases, DB format changes, scanner CVEs, adapter bug fixes. Today none of that can reach users without cutting a Harbor release. The pin in `versions.env` is currently a release behind the scanner's latest tag, which is the normal steady state of this arrangement, not an oversight.

**Ship on demand.** A scanner fix becomes a chart value bump or an image tag change, available the same day it is released, instead of waiting for the next Harbor release train.

**Harbor releases get smaller and faster.** Harbor releases stop carrying scanner changes, so their scope and risk shrink. Conversely, a scanner regression no longer blocks or gets bundled into a Harbor release.

**Clear ownership and provenance.** The scanner is developed, tested, reviewed and released in the repository that owns it. Its images are signed and carry SBOM attestations from that pipeline. Version numbers mean what they say: the image tag identifies the scanner release, not the Harbor release it happened to be built alongside.

**Less CI work and fewer moving parts in harbor-next.** Today any change under `src/**` rebuilds the adapter image for every platform even though nothing in `src/**` can affect it. Removing it drops a component from every image matrix, three version pins, two source clones and one dockerfile from this repository.

[...]

PR #893, opened 2026-09-14
chore(deps): Bump the pinned dependencies of the vendored builds

bupd:
Three of the binaries we ship are compiled from a third-party tree that
`taskfile/build.yml` clones and then pushes past a CVE with `go mod edit -require`.
Nothing re-checks those pins, and every one of them has now been overtaken by
upstream — two to the point of pulling the dependency *backwards*.

| Build | Pin before | Upstream ships | Effect today |
|---|---|---|---|
| `binary:lprobe` | `x/net@v0.55.0` | v0.2.1 → 0.58.0 | **downgrade**, re-opens CVE-2026-46600 |
| `binary:lprobe` | `go-jose/v4@v4.1.4` | v0.2.1 → 4.1.4 | no-op |
| `binary:trivy` | `containerd/v2@v2.3.2` | v0.74.0 → 2.3.3 | **downgrade**, and 2.3.2 carries CVE-2026-53495 |
| `binary:registry` | `x/crypto@v0.52.0` | distribution v3.1.1 → 0.49.0 | still an upgrade, but 0.52.0 is itself vulnerable |
| `binary:registry` | `x/net@v0.55.0` | distribution v3.1.1 → 0.52.0 | same |

The published `v2.15.8` images show exactly these forced versions — x/net 0.55.0 in
lprobe, x/crypto 0.52.0 in the registry binary, containerd 2.3.2 in the Trivy CLI —
so this is what we are actually shipping, not a theoretical risk.

## Changes

**lprobe → 0.2.1, both pins dropped.** v0.2.1 already carries everything the scan
asks for (grpc 1.83.2, x/net 0.58.0, x/text 0.41.0, go-jose 4.1.4). Keeping the
x/net pin would drag 0.58.0 back to 0.55.0, so the bump only works with the pins
removed. Closes CVE-2026-46600, CVE-2026-56852, CVE-2026-84303, CVE-2026-84304,
CVE-2026-84445 and GHSA-hrxh-6v49-42gf.

lprobe is copied into all seven images, so its findings are counted seven times:
**42 of the 402 findings** across the fleet.

**Trivy CLI: containerd → v2.3.5.** Closes CVE-2026-53495. This is the vendored CLI
we compile, not `TRIVY_BASE_IMAGE_VERSION` — the scanner version users see is
unchanged, and the adapter still reports Trivy 0.74.0.

**Registry: pins moved to current patches, x/text and grpc added.** distribution
v3.1.1 resolves x/crypto 0.49.0, x/net 0.52.0 and x/text 0.35.0 on its own, so these
stay as forced upgrades — just current ones. Closes CVE-2026-56854, CVE-2026-56855,
CVE-2026-78662, CVE-2026-46600, CVE-2026-56852 and the four grpc advisories.

## Verification

Each binary was rebuilt and rescanned with Trivy 0.74.0:

| Binary | Before | After | What is left |
|---|---:|---:|---|
| `lprobe` | 14 | **0** | — |
| `registry` | 18 | **1** | GO-2026-5932, no published fix |
| `trivy` | 34 | **6** | Trivy's own graph — see #641 |

`go build ./...` is clean on the full commercial megamerge, and the stack runs
locally with every component healthy, including the trivy-adapter whose healthcheck
is the rebuilt lprobe.

Refs #891


bupd:
Agreed on the substance — a one-time refresh does not stop it happening again, and that is the most interesting thing this PR turned up.

Keeping it out of this PR deliberately, for two reasons: it is a different concern from the refresh itself (and a reviewer should be able to check the version table here without also reviewing new build logic), and the brief this came from says to propose a guard rather than enforce one unasked.

Concrete design, recorded in #891:

Make the pin application self-verifying inside the build task rather than adding a CI gate. Before applying a pin, read what the cloned tree resolves on its own (`go list -m -f {{.Version}} <module>`); if that is already at or above the pin, the pin is redundant or a downgrade, and the build fails telling you to raise or drop it. That runs everywhere the binary is built, CI included, needs no workflow change, and would have caught both the lprobe x/net and the Trivy containerd downgrades the day the upstream version moved past them.

The severity-threshold scanning idea in #891 is the other half, and is the part that genuinely should not be a blocking gate — a new `affected` CRITICAL in someone else's base image is not something a PR author can act on.

Happy to open that as a follow-up PR if you want it before this one lands.

bupd:
Extended: `binary:trivy` now also pins x/crypto v0.56.0, grpc v1.83.2 and moby/go-archive v0.3.0 — main is already on trivy 0.74.0, which still ships those vulnerable (#901). Same pin set as the release-2.15 counterpart in #898, build-verified locally.

PR #955, opened 2026-09-19
fix: prevent project ListAll goroutine leak

### LINKED ISSUE #954: [Bug]: Goroutine leak in project.ListAll when consumer stops reading
### Description

`ListAll` in `src/controller/project/util.go` spawns a background goroutine that sends project results on a buffered channel. All channel sends are bare — no cancellation check:
```go
for _, p := range projects {
    ch <- Result{Data: p}
}
```
If a caller stops consuming the channel before the producer finishes, the producer goroutine can block on a channel send indefinitely. The blocked goroutine and objects reachable from it remain in memory indefinitely.

There are two conditions required to prevent the leak, and neither is met:

1. Producer side (util.go): Channel sends have no select on ctx.Done(), so even if the context is cancelled, the goroutine cannot observe it while blocked on a send.
2. Caller side (garbage_collection.go): markOrSweepUntaggedBlobs passes ctx.SystemContext() directly to ListAll. The system context is long-lived and never cancelled. When GC exits early via return nil, errGcStop, nobody cancels the context, so the producer has no signal to stop.

### Steps to Reproduce

1. Trigger a garbage collection job with a large number of projects (e.g. 20+ projects, chunk size 2).
2. While GC is iterating over projects via ListAll, send a stop signal to the GC job (OPCommand.IsStop() == true).
3. GC exits markOrSweepUntaggedBlobs via return nil, errGcStop and stops reading from the ListAll channel.
4. The ListAll producer goroutine blocks on ch <- Result{Data: p} once the channel buffer fills.
5. Since ctx.SystemContext() is never cancelled, the goroutine remains blocked indefinitely.

### Expected Behavior

When a caller stops consuming the ListAll channel (e.g. GC receives a stop signal), the producer goroutine should detect the cancellation and exit cleanly, releasing the goroutine and associated memory.

### Actual Behavior

The producer goroutine blocks on a bare channel send (ch <- Result{Data: p}) indefinitely. The goroutine and all objects reachable from it remain in memory for the lifetime of the process.

### Version

main (current HEAD)

### Environment

Local development

### Relevant Logs

```shell
No explicit error logs are produced. The leak is silent — the goroutine blocks without logging. It can be observed via runtime.NumGoroutine() or pprof goroutine dumps showing a goroutine stuck at:

goroutine [chan send]: github.com/goharbor/harbor/src/controller/project.ListAll.func1() src/controller/project/util.go:49
```

### Checklist

- [x] I have searched for similar issues and this is not a duplicate
- [x] I have provided all the requested information

khuswant18:
## Summary

Fixes a goroutine leak in `project.ListAll` (`src/controller/project/util.go`) that occurred when a consumer stopped reading from the channel early without the producer noticing context cancellation:

1. **Context cancellation check in `project.ListAll`**:
   - Clones and configures `query` before launching the goroutine, resolving a data race between caller stack references and the producer.
   - Adds `case <-ctx.Done(): return` check at the start of the pagination loop.
   - Encloses channel sends (`ch <- Result{...}`) inside `select` statements with `case <-ctx.Done(): return` to prevent the background producer goroutine from blocking indefinitely if the consumer stops draining `ch`.

2. **Caller context lifecycle in GC**:
   - In `GarbageCollector.markOrSweepUntaggedBlobs` (`src/jobservice/job/impl/gc/garbage_collection.go`), wraps `ctx.SystemContext()` with `context.WithCancel()` and `defer cancel()`. When the GC job halts early (e.g. `gc.shouldStop(ctx)`), the context is cancelled immediately, releasing the background producer.

3. **Unit Tests**:
   - Added `src/controller/project/util_test.go` covering normal pagination, context cancellation on early consumer exit, pre-cancelled context, error handling, and empty results.
   - Added `src/jobservice/job/impl/gc/mark_sweep_test.go` verifying context cancellation terminates the producer on early GC stop.

## Related Issues

Fixes #954

## Type of Change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Documentation (`docs:`)
- [ ] Tests (`test:`)

## Release Notes

Fixed a goroutine leak in project listing when consumers terminate consumption early.

## Testing

Ran unit tests with `-race`:
```bash
go test -race -v -count=1 ./controller/project/...
go test -race -v -count=1 -run TestMarkOrSweepUntaggedBlobs_ContextCancelledOnStop ./jobservice/job/impl/gc/...
go vet ./controller/project/... ./jobservice/job/impl/gc/...
task test:quick
```
The regression tests were also run against the original implementation and reproduced the goroutine leak before the fix.

Checklist
 - [x] Commits are signed off (git commit -s)


khuswant18:
@bupd I’ve updated the description with the before/after changes. Please let me know if there’s anything else.

